PR #21, opened 2015-03-12
Don't use the word 'manager'

stefwalter:
Cockpit is not a management system. It's an admin interface.


dperpeet:
as discussed on irc, I think this phrasing is much better than the old one!

PR #27, opened 2015-06-02
add instructions for atomic-run command on Fedora Atomic

andreasn:


petervo:
The container is really only for connecting to that host directly with your browser. You can still add it to another cockpit without doing that. Is it worth explaining that a little more?

andreasn:
Yeah, that would be good to add.
Something like:
"Connect to a Fedora Atomic host from an existing host running Cockpit with the Add Server UI.
Alternatively you can install Cockpit on the Atomic host with the command <code>atomic run cockpit/ws</code>"

running.html
```diff
@@ -82,7 +82,8 @@
           </div>
           <div class="col-md-12 os-info os-info-fedoraatomic" style="display: none;">
             <div class="os-infobox">
-            Cockpit comes installed by default in Fedora Atomic.
+            Connect to a Fedora Atomic host from an existing host running Cockpit with the Add Server UI.<br/>
```
stefwalter: Suggest using the word 'using' instead of 'with' here. Makes it clearer.

running.html
```diff
@@ -82,7 +82,8 @@
...
+            Alternatively you can install Cockpit on the Atomic host with the command <code>atomic run cockpit/ws</code>
```
stefwalter: You need a ```sudo``` in the beginning of that command.

cgwalters:
See also https://fedorahosted.org/cloud/ticket/105

PR #37, opened 2015-10-29
Updated screenshots for frontpage

### LINKED ISSUE #32: update screenshots
uses old navigation

Akasurde:
Fixes #32 
Signed-off-by: Abhijeet Kasurde akasurde@redhat.com


stefwalter:
CC @andreasn 

Thanks for doing this @Akasurde !

But I think the images need to be taken with a smaller browser window, so that the contents are legible. For reference the terminal should take up most of the right-hand side view in such a screenshot.

andreasn:
I agree with Stef's comment.
I also think it would be good to show another page than the terminal on the "Easy to use" screenshot. Reason is that a system with Cockpit on it, where someone is running the terminal is exactly as easy to use as a machine with just the console and no Cockpit, so hence the screenshot doesn't carry the idea of the text as well.
If you're able to take one of the Storage page that would be great, since that's usually one of the big pain points of using the CLI-tools on most servers (plus that it's very visual with graphs, icons and such).

Akasurde:
ok, I will try to come up with new screenshots as discussed in comments.

PR #40, opened 2016-01-19
Add Debian to running.html

mvollmer:
Issues:
- Logo doesn't have transparent background.
- Layout of logo row is messed up.


larsu:
I've implemented the new design and added debian in #42.

PR #41, opened 2016-02-02
Fix the documentation for centos atomic host

mscherer:
While preparing a demo for devconf, we found that Centos Atomic got updated and do the same as Fedora.


stefwalter:
Thanks. Merged.

PR #42, opened 2016-03-23
Redesign running.html

larsu:
Implement redesign by @andreasn.

Add debian instructions.

Allow toggling info boxes (doesn't make a lot of sense, but it felt weird to me that this didn't work).


andreasn:
Just to confirm. Is the order tested OSes in stable releases, in alphabetical order and then tested OSes, but unstable releases?

The Arch and Ubuntu links seem a bit big. Especially the arch one.

andreasn:
Screenshot for reference

screenshot from 2016-03-23 13-31-04

andreasn:
but maybe the Ubuntu logo size is OK, but the Arch one certainly feels biggest of them all on the whole page.

andreasn:
It would be nice with some link to more info about our testing framework possibly

larsu:
> Just to confirm. Is the order tested OSes in stable releases, in alphabetical order and then tested OSes, but unstable releases?

Yes. Seemed most logical to me.

> [...] but the Arch one certainly feels biggest of them all on the whole page.

Yeah. I sized the logos so that the logotypes are similar in size and baseline. Arch's logo has a huge icon to type ratio (and their policy doesn't allow changing it). I don't like how it looks when I resize it, because it looks unaligned to me. Do you prefer this version?

image

>  It would be nice with some link to more info about our testing framework possibly

Do we have anything other than HACKING.md in the repository?

andreasn:
> Yes. Seemed most logical to me.
I agree.

larsu:
I've updated the archlinux logo.

andreasn:
I modified the size of the arch logo here 33c9058aed004dca2d4c67df4bd123d9ce6d79a4
What do you think?

andreasn:
Anyway, with or without that. I think this looks go to go in!

PR #43, opened 2016-03-25
fix install command on Debian

Smana:


larsu:
I intentionally chose `apt` here because that seems to be the preferred tool these days and it's what the debian administrator's handbook uses in its examples [1].  

Why did you change it?

[1] https://debian-handbook.info/browse/stable/sect.apt-get.html

Smana:
Yes sorry, i'm not up to date :p

stefwalter:
So does 'apt' work on all Debians that Cockpit is targetting? If so, I'm certainly not against changing it back.

larsu:
Nah, it doesn't matter. I was just curious if there was a specific reason for the change.

PR #46, opened 2016-05-24
running: Change apt-get source to amd64 for Debian

dperpeet:
On the "running Cockpit" page, change apt source line to prevent
apt-get from trying to fetch packages for other architectures.

### DIFF
--- a/running.html
+++ b/running.html
@@ -97,7 +97,7 @@
             <div class="os-infobox">
               Cockpit is not part of the official Debian repositories, but you can add a repository which contains weekly builds:<br>
               <ol>
-                <li>Add the line <code>deb https://fedorapeople.org/groups/cockpit/debian unstable main</code> to <code>/etc/apt/sources.list</code></li>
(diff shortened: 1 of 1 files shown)


larsu:
Thanks!

Could you please also change the line right above it to: "[...] contains weekly builds for amd64"?

larsu:
Alternatively, we could export an empty i386 index: https://github.com/larsu/cockpituous/commit/ac957baee8e0a5078bdfdbbfaa50bea88e1f7e8b

This way, people won't have to change anything. What do you think?

dperpeet:
needs a rebase, what is the best way to do this?

larskarlitski:
> needs a rebase, what is the best way to do this?

I still think we should fix the repository instead of making it more complicated for people to add it. I've rebased my commit and opened [#53 for cockpituous](https://github.com/cockpit-project/cockpituous/pull/53).

stefwalter:
@larskarlitski I don't think we need this anymore right?

larskarlitski:
> @larskarlitski I don't think we need this anymore right?

No, it was fixed with the commit I linked above.

PR #49, opened 2016-11-14
Add Ubuntu to the main list of operating systems

stefwalter:


stefwalter:
@andreasn I need  help with the layout here. How can I make this look better at big sizes. All the icons seem much smaller.

andreasn:
Will take a look

andreasn:
Maybe something like this instead?
I gave the top 3 items the class col-sm-4 and the two bottom ones col-sm-6

screenshot from 2016-11-15 14-28-04

running.html
```diff
@@ -128,6 +128,18 @@
               </ol>
             </div>
           </div>
+	 </div>
+
+        <div class="row">
+          <div class="col-md-12"><div class="browser-header"><p>Stable, tested, delivered in a repository:</p></div></div>
```
andreasn: Can probably remove the colon here

running.html
```diff
@@ -170,10 +182,8 @@
           </div>
         </div>
 
-        <hr>
-
         <div class="row">
-          <div class="col-md-12"><p>Less tested and/or irregularly updated packages are available on</p></div>
+          <div class="col-md-12"><div class="browser-header"><p>Less tested and irregularly updated:</p></div></div>
```
andreasn: And here too

PR #59, opened 2017-03-02
Release 133

dperpeet:
We can't merge this yet, since the [fedora packages](https://bodhi.fedoraproject.org/updates/FEDORA-2017-ff19c0da94) are still pending


dperpeet:
Recreated blog directory based on latest repo changes, content is the same.

Issue #69, opened 2017-06-14
Guide: missing option for cockpit.file


machadotiago:
In the Cockpit Guide at the cockpit.file session ( http://cockpit-project.org/guide/latest/cockpit-file.html ) it's mentioned that only those options are available for the file function
```
cockpit.file(path,
                    { syntax: syntax_object,
                      binary: boolean,
                      max_read_size: int,
                    })
```

As far as I could see from the sources when "file" is called the options passed to the cockpit channel just extends the options send through the file function. This means that some other options can be passed like:

```
host
superuser
```
I think, in this case "host" is not so useful but "superuser" is very useful...specially in cases where you want to save/replace a file that requires superuser access.
This took me a while to find out (since it's not in the Guide) so, I created a pull request adding this info to the guide page.

machadotiago:
Closing as dicussed here: https://github.com/cockpit-project/cockpit-project.github.io/pull/70

PR #70, opened 2017-06-14
Adding undocumented option to cockpit.file

machadotiago:
It's also possible to use superuser with cockpit.file but it was not
documented in the Cockpit Guide as mentioned here: https://github.com/cockpit-project/cockpit-project.github.io/issues/69

### DIFF
--- a/guide/latest/cockpit-file.html
+++ b/guide/latest/cockpit-file.html
@@ -38,6 +38,7 @@ file = cockpit.file(path,
                     { syntax: syntax_object,
                       binary: boolean,
                       max_read_size: int,
+                      superuser: string
                     })
 
 promise = file.read()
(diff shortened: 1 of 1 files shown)


mvollmer:
Thanks!

This repo is unfortunately not the place to fix this.  The sources for the cockpit guide are here: https://github.com/cockpit-project/cockpit/blob/master/doc/guide/cockpit-file.xml

In fact, all options accepted by `cockpit.channel` are also possible with `cockpit.file`.  The interesting ones are `superuser` and `host`.  (`payload` will be determined by `cockpit.file`, of course, and `binary` is already documented.)

Could you open a pull request for https://github.com/cockpit-project/cockpit?

mvollmer:
> In fact, all options accepted by `cockpit.channel` are also possible with `cockpit.file`.  The interesting ones are `superuser` and `host`.  (`payload` will be determined by `cockpit.file`, of course, and `binary` is already documented.)

(Ahh, you knew this. Sorry for the noise.)

guide/latest/cockpit-file.html
```diff
@@ -84,6 +85,12 @@
 <p>It is not an error when the file does not exist. In this case, the
       <code class="code">done()</code> callback will be called with a <code class="code">null</code>
       value for <code class="code">content</code> and <code class="code">tag</code> is <code class="code">"-"</code>.</p>
+<p>Set <code class="code">superuser</code> to "require" for reading or writing a file as root instead of the logged
```
mvollmer: I think we should just say that `superuser` is also allowed and link to the docs for `cockpit.channel`.

machadotiago: fair enough, I will change the text and make the pull request to the correct repo. 
Thanks !

PR #72, opened 2017-06-21
ideals: Add "Cockpit is tested"

martinpitt:
This came up a few times, and last night I just remembered it again. So here it is :-)

### DIFF
--- a/ideals.html
+++ b/ideals.html
@@ -151,6 +151,12 @@
 	<p>Cockpit allows others to add additional <span class="caps">UI</span> pieces. We do
 	not allow arbitrary extension of the <span class="caps">UI</span> all over the place, but
 	at distinct extension&nbsp;points.</p>
+
+	<h2>Cockpit is&nbsp;tested</h2>
+	<p>Every feature and change in Cockpit is covered by an automatic test.
+	A proposed change only lands in the code base after it passes all
+	tests. Tests run on all <a href="running.html">Operating System</a>
+	releases on which Cockpit is supported.</p>
 </div>
 
 </article>


ideals.html
```diff
@@ -151,6 +151,12 @@
 	<p>Cockpit allows others to add additional <span class="caps">UI</span> pieces. We do
 	not allow arbitrary extension of the <span class="caps">UI</span> all over the place, but
 	at distinct extension&nbsp;points.</p>
+
+	<h2>Cockpit is&nbsp;tested</h2>
+	<p>Every feature and change in Cockpit is covered by an automatic test.
+	A proposed change only lands in the code base after it passes all
+	tests. Tests run on all Operating System releases on which Cockpit is
```
dperpeet: These "all OS releases" could be a link to the actual page.

ideals.html
```diff
@@ -151,6 +151,12 @@
 	<p>Cockpit allows others to add additional <span class="caps">UI</span> pieces. We do
 	not allow arbitrary extension of the <span class="caps">UI</span> all over the place, but
 	at distinct extension&nbsp;points.</p>
+
+	<h2>Cockpit is&nbsp;tested</h2>
+	<p>Every feature and change in Cockpit is covered by an automatic test.
```
dperpeet: Do we meet those strict criteria? "Cockpit features and changes are covered by automatic tests."?

ideals.html
martinpitt: I made "all OS" a link to http://cockpit-project.org/running.html, but we don't have an easily accessible per-release list (at most in our bots/tests-scan, but I don't think we want to link to that).

ideals.html
martinpitt: We certainly are far from 100% coverage, but in the spirit of ["Ideals are like stars..."](https://www.brainyquote.com/quotes/quotes/c/carlschurz389236.html) we generally try to abide by it and reject new features/fixes without tests.

stefwalter: I think this is good. Features that are not tested are basically proof of concepts, or failures. They rot and don't last. In broad terms: anything in Cockpit which hasn't rotted has some form of test covering it.

PR #76, opened 2017-07-21
Cockpit 146

martinpitt:
Release is done already: https://github.com/cockpit-project/cockpit/releases/tag/146


content/cockpit-146.md
```diff
@@ -0,0 +1,65 @@
...
+summary: Cockpit with Software Updates improvements and GCE support
+comments: true
+---
+
+Cockpit is the [modern Linux admin interface](http://cockpit-project.org/). We release regularly.
+Here are the release notes from version 146.
+
+### Software Updates page shows more information
+
+The "Available Updates" and "Restart recommended" pages now show the packages
```
dperpeet: Do we want to use quotes (`"`) now instead of ```?

martinpitt: IMHO that looks rather poor:

verb

But I'm happy to drop the quotes entirely, at least for page titles.

martinpitt: In previous blogs we never quoted the Page Titles at all, and referred to UI elements with ``. Updated accordingly. Thanks for spotting!

PR #81, opened 2017-09-05
Running: Re-add install instructions for Debian 9

martinpitt:
Port commit 6379b08 over to the new Jekyll website, as the conversion
branched off before that commit.

Add explicit line breaks and \ to the commands to avoid line breaks.

Rename the capture labels to something more suitable.

### DIFF
--- a/running.md
+++ b/running.md
@@ -104,24 +104,32 @@ sudo firewall-cmd --reload
 {% endcapture %}
 
 
-{% capture debian-unstable %}
-Cockpit is included in Debian unstable.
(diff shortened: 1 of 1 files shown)


martinpitt:
Due to the way github works, this is now a "new" PR for re-fixing the website after merging the Jekyll transition. You can see the result on https://martinpitt.github.io/cockpit-project.github.io/running.html#debian which puts back the Debian 9 instructions which got lost from http://cockpit-project.org/running.html#debian

PR #83, opened 2017-09-11
Port to Jekyll

garrett:
Here's the Jekyll port of the website, reformatted as branches on top of the Pelican site, so it can be implemented as a pull request.


garrett:
This PR shouldn't be rebased or squashed — it should be directly merged.

I'll test it from the command line to see what happens (and if I need to work on the git merging of separate histories more).

PR #85, opened 2017-09-12
Running: Update install instructions for Debian 8

martinpitt:
Cockpit is now in official jessie-backports-sloppy, for old stable 8.x.
This is built for all architectures, so in particular this can now be
used on Raspian (see https://github.com/cockpit-project/cockpit/issues/6560).

### DIFF
--- a/running.md
+++ b/running.md
@@ -105,7 +105,7 @@ sudo firewall-cmd --reload
 
 
 {% capture debian-official %}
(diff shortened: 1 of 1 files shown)


martinpitt:
Result can be seen at https://martinpitt.github.io/cockpit-project.github.io/running.html#debian

Issue #52, opened 2016-12-02
Add Atom/RSS-feed


haeckle:
Would be wonderful to follow the blog with my feedreader.

http://docs.getpelican.com/en/stable/settings.html#feed-settings

garrett:
There's now a page that lists feeds, located at http://cockpit-project.org/blog/feeds/

PR #89, opened 2017-09-14
Atom page

garrett:
Adds a page for displaying Atom feeds. 

RSS is specifically mentioned so if people search for "atom", "RSS", or "feed", this page should show up.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -4,8 +4,14 @@
 - title: Blog
   url: /blog/
 
+- title: Try it out
+  url: /running.html
+
 - title: Contribute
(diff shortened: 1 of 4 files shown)


garrett:
...I'm not sure why GitHub is showing the 'reworked nav' commit here. It's already included in master and not part of the branch locally.

When I do a `git diff origin/master`, it just shows the Atom commit. 🙄

garrett:
I used the "step 1" instructions to merge it locally to test, and it works. It only includes the 1 commit, so the rebase and merge button should make it work fine.

garrett:
This fixes #52, by-the-way. (Feeds existed, but now they're exposed a bit better.)

PR #100, opened 2017-09-29
Gem update to GitHub Pages version 161 (2017-09-29)

garrett:

PR #105, opened 2017-10-10
updated gems to GitHub Pages 163 (2017-10-10)

garrett:


garrett:
Obsoletes #100

martinpitt:
@garrett: OOI, does that have any practical effect? I thought our pages were rendered by GitHub. Or is this only for rendering them locally?

garrett:
Yep, it's built on GitHub, but it's still good to bump up the gems used to match GitHub's, to make sure there are no surprises when it's deployed on GitHub versus when we're building and testing locally.

PR #117, opened 2017-11-23
running: Fix Atomic install instructions

### LINKED ISSUE #116: Install instructions for Cockpit on Atomic doesn't work
On my Fedora Atomic 27 system, `sudo atomic run cockpit/ws` gave me:
`The image 'ws' appears to have not been installed and has an INSTALL label.  You should install this image first.  Re-run with --ignore to bypass this error.`
And Cockpit didn't start.

Instead I had to run:
`sudo atomic install docker.io/cockpit/ws`
`sudo atomic run docker.io/cockpit/ws`

@martinpitt verified this was the case on his Fedora 26 machine as well.

martinpitt:
`atomic install` is explicitly necessary, otherwise it does not get
properly installed.

Fixes #116

### DIFF
--- a/running.md
+++ b/running.md
@@ -80,6 +80,7 @@ Alternatively you can access Cockpit directly on the Atomic Host if SSH password
 
 1. Run the Cockpit web service container: 
 ```
+sudo atomic install cockpit/ws
 sudo atomic run cockpit/ws
 ```
 {% endcapture %}


martinpitt:
Preview: https://martinpitt.github.io/cockpit-project.github.io/running.html#atomic

Issue #51, opened 2016-11-21
Running: contrast of section headers is too low


larskarlitski:
The W3C [recommends][1] a contrast ratio of 4.5:1 for most text, so that it is also readable for people who have various forms of vision deficiencies.

There's a tool to check if the contrast ratio is high enough: http://webaim.org/resources/contrastchecker/?fcolor=b4b7af&bcolor=ffffff

[1]: https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

garrett:
Fixed in #128.

PR #128, opened 2018-01-18
accessibility improvements for contrast, alt text, search, blog, etc.

garrett:
Commit 06e54f5 fixes issue #51.

The rest of the commits were a result of doing accessibility audits (in Chrome) across the site.


garrett:
Don't merge yet — I'm looking through to see what we can improve on the guide too.

garrett:
While these are accessibility related, the changes span accross the site. As such, they should probably be rebased instead of squashed.

PR #136, opened 2018-02-07
Cockpit 161

martinpitt:
Rendered preview: https://martinpitt.github.io/cockpit-project.github.io/blog/cockpit-161.html


martinpitt:
Argh, wrong screenshot for hw info - fixed.

_posts/2018-02-07-cockpit-161.md
```diff
@@ -0,0 +1,86 @@
...
+
+![Machines Create new VM](/images/machines-create-iso.png)
+
+Thanks to Dominik Perpeet and suomiy for this feature!
+
+### VMs running in Kubernetes can now be deleted
+
+[Cockpit 160](http://cockpit-project.org/blog/cockpit-160.html) added an
+initial overview of kubevirt Virtual Machines. This page now got a first action
+to delete a running VM.
```
stefwalter: Is this really only VMs that are running? Or also stopped VMs?

_posts/2018-02-07-cockpit-161.md
```diff
@@ -0,0 +1,86 @@
...
+future versions.
+
+![System Hardware Info](/images/system-hwinfo.png)
+
+### Load Application metadata (Appstream) packages on demand on Debian/Ubuntu
+
+[Cockpit 160](http://cockpit-project.org/blog/cockpit-160.html) started to
+install the [AppStream metadata](https://www.freedesktop.org/wiki/Distributions/AppStream/)
+on demand on the Apps page on Fedora. Debian, Ubuntu, and derivatives use a
+different mechanism for that, which is now being handled by Cockpit.
```
stefwalter: Still no real applications listed here on Debian/Ubuntu right? If so, is this even worth mentioning? 

Until we have real stuff listed, this is a non-feature from a user point of view. Please correct me if I'm wrong and we have real usable applications listed.

_posts/2018-02-07-cockpit-161.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Rename cockpit-ovirt package to cockpit-machines-ovirt
+
+The previous package name collided with a package that is already available in
+Red Hat Virtualization Host. This only affects Fedora, as other operating systems
+do not ship the Cockpit ovirt extension.
```
stefwalter: Someone at FOSDEM said that CentOS erroneously ships this ... and makes it a required subpackage. Leading to annoyances of installing OVirt client tooling when installing Cockpit.

_posts/2018-02-07-cockpit-161.md
martinpitt: In current kubevirt there is no concept of a "stopped" VM, as they are ephemeral (much like a pod). So "running" and "not existing" are the only two states. AFAIK the kubevirt devs want to introduce a concept of "offline VMs" at some point, but it doesn't exist right now.

_posts/2018-02-07-cockpit-161.md
martinpitt: You are right. sosreport isn't a thing on Debian and Fleet commander etc. neither. So I removed it.

_posts/2018-02-07-cockpit-161.md
martinpitt: I see how CentOS 7 could ship the cockpit-ovirt package, but our packaging (tools/cockpit.spec) never made any of cockpit-{machines,ovirt,machines-ovirt} a dependency of the `cockpit` metapackage, and we never actually packaged it in the RHEL downstream dist-gits. Also, -ovirt is fairly new, so it's curious that it even already found its way there. Doesn't CentOS use or at least base on the rhel dist-gits? Maybe the CentOS packagers did do a deviation there.

I'm downloading the centos-7 image to verify.

martinpitt: On current CentOS 7, `yum search cockpit` does not include ovirt, and installing `cockpit` only pulls in -bridge, -system, and -ws. So as far as I can see, we should be safe here. Maybe that someone used some third-party repo?

martinpitt:
Release is done, so can you please re-review?

_posts/2018-02-07-cockpit-161.md
mareklibra: I can remember there was issue in installing cockpit-ovirt and one another package by default, but this was fixed round Cockpit 155

PR #137, opened 2018-02-08
ARIA landmarks

garrett:
Added several ARIA landmarks around the site to make keyboard navigation and screen reading better.

### DIFF
--- a/_layouts/guide.html
+++ b/_layouts/guide.html
@@ -8,8 +8,15 @@ layout:
   {% assign head_tag = '</head>' %}
   {% assign nav_tag = '<table class="navigation"' %}
   {% assign book_tag = '<div class="book"'%}
+  {% assign toc_tag = '<div class="toc"'%}
+  {% assign part_tag = '<div class="part"'%}
+  {% assign ref_tag = '<div class="refentry"'%}
+  {% assign chap_tag = '<div class="chapter"'%}
 
   {% assign html_mod = '<html class="cockpit-guide" lang="en"'%}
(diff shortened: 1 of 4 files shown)


garrett:
I'm practicing accessibility features on the website as a warm-up for Cockpit itself. The website is much less complex. :wink:

The accessibility meeting videocall talked about screenreader usage between using ChromeVox and checking the site with aXe, I realized we're lacking landmarks. Thus, I added landmarks. :smile:

_layouts/guide.html
```diff
@@ -31,15 +38,24 @@
...
 
 {%
   endcapture
 %}{{ content
   | replace_first: html_tag, html_mod
   | replace_first: head_tag, head_mod
+  | replace_first: ref_tag, ref_mod
+  | replace_first: chap_tag, chap_mod
+  | replace_first: part_tag, part_mod
+  | replace_first: toc_tag, toc_mod
```
garrett: All the above changes in this guide rewriting file are a bit hackish.

Ideally, it would probably check for the existence of a tag before replacing it. It would be more involved, however, and I checked to ensure this works.

(In this implementation, it replaces the tags with mods, which may be modified versions or not. If values are blank, it just replaces blank with blank. I verified it works for all variants of the guide pages.)

garrett: Note: Liquid (and nearly everything you have to do in Liquid) is hackish due to constraints... but it's Jekyll's templating language so it is what it is.

PR #141, opened 2018-02-28
running: bumped up browser requirements to IE11

garrett:
- IE10 is completely dead, even according to Microsoft. 
- We use several features that are really dodgy, if supported at all in IE10.
- IE11 is the last version, so there's no `+` anymore.


garrett:
More info on death of IE:
- https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support
- https://support.microsoft.com/en-us/help/17454/lifecycle-faq-internet-explorer

garrett:
:balloon: :tada: :balloon:

PR #147, opened 2018-03-14
Proposal: update ideals

larskarlitski:
Like we discussed, let's update Cockpit's ideals.

Instead of splitting them into user-facing and team-facing, I've ordered them from user-to-team, because there's no clear delineation between the two.

This is only a proposal with which I want to start the discussion.

### The commit message contains an overview of the changes:

Cockpit has been around for a while now and some of its ideals were out
of date or overlapping with others, some are simply features of the
software, and we developed a new one.

In "is discoverable", remove the sentence that says we're not there yet.
We're still not there, but much closer than we were when this was
written.

Remove "remotable". It's very obvious and even stated that it was.

Merge "does not own the server", "does not store data or policy", and
"reacts to the server" into "Cockpit directly interacts with and reacts
to the system". This removes a lot of overlap and brings down the number
of ideals.

Remove the bits about "pretty UI" from "zero footprint", because that's
just selling Cockpit short.

Remove "firewall friendly" and "pluggable", as those are established
features now.

Add usability testing to "cockpit is tested".

Also sort them differently. Roughly start with the ones that deal with
the qualities of Cockpit itself towards the qualities of making it.

Add a new one about being backwards-compatible and trying not to add new
APIs.


ideals.md
```diff
@@ -10,75 +10,52 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
-### Cockpit reacts to the server
-
-Cockpit dynamically updates itself to reflect the current state of the server, within a time frame of a few seconds. This is important to the admin experience. An admin should feel that they are on the server itself, and not in some tool. The admin should be able to effectively use a terminal, Cockpit, and other management tools all concurrently.
+### Cockpit is only the project name
 
-### Cockpit is firewall friendly
+People using Cockpit should feel they are interacting with the underlying Server OS. Cockpit can look different on different operating systems, because it’s the UI for the OS, and not a external tool.
 
-Cockpit opens one port for browser connections: by default that is 9090. We work well in cases where a bastion host runs Cockpit and connects out to the other servers via SSH.
+When packaging a stable release of Cockpit for release with a given OS, most of the “Cockpit” branding should be removed, and OS specific branding is put in its place.
```
andreasn: Might worth mentioning that we minimize the use of the term "cockpit" in the UI as much as possible, that it's referred to as "Web Console" (or whatever term we use, I think it's slightly different in different places :/ ), and that a lot of the branding is taken care of automatically, such as parsing /etc/os-release for the product name, loading the OS logo from /usr/share/logos etc.

ideals.md
```diff
@@ -10,75 +10,52 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
 ### Cockpit is discoverable
 
-It should be possible to discover how to perform tasks through Cockpit without reading a help file. Astute readers may note that we have not completely acheived this yet, but this drives design decisions for Cockpit.
+It should be possible to discover how to perform tasks through Cockpit without reading a help file.
 
-### Cockpit is remotable
+### Cockpit directly interacts with and reacts to the system
 
-Cockpit can be effectively used when you are not sitting next to the server. This may seem obvious. Well, it is obvious.
+Cockpit does not store its own opinion of the state of the server anywhere. The UI always reflect the actual current state of the server and reacts to changes immediately, using the same system APIs that are used by command line tools.
```
stefwalter: This is a runon sentence. Lets separate it. ```s/the server and reacts to changes/the server. Cockpit reacts to system changes/```

ideals.md
```diff
@@ -10,75 +10,52 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
 
 ### Cockpit is tested
 
 Every feature and change in Cockpit is covered by an automatic test. A proposed change only lands in the code base after it passes all tests. Tests run on all [Operating System](running.html) releases on which Cockpit is supported.
 
+In addition, the design of new features and changes in existing features are tested in usability studies.
+
+### Cockpit is backwards-compatible
+
+Cockpit can connect to other instances of itself. These might be much older, because we support many operating systems with different release cadences. It is important that this keeps working, so we try to as little API as possible to `cockpit.js` and no other API.
```
stefwalter: Lets also add:

Cockpit's Javascript API in ```cockpit.js``` may not change in an incompatible way. External Cockpit components (or plugins) that use the ```cockpit.js``` javascript API should continue to function despite newer versions of Cockpit being released. 

There is no forseen major incompatible version change for this Javascript API in Cockpit's future.

ideals.md
```diff
@@ -10,75 +10,52 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
 
 Cockpit may expose UI elements that interact with configuration management style software installed on the server. Such software may have a declarative configuration, as one would expect. But Cockpit itself has no concept of configuration management, or a desired state for the server.
 
-### Cockpit reacts to the server
-
-Cockpit dynamically updates itself to reflect the current state of the server, within a time frame of a few seconds. This is important to the admin experience. An admin should feel that they are on the server itself, and not in some tool. The admin should be able to effectively use a terminal, Cockpit, and other management tools all concurrently.
+### Cockpit is only the project name
 
-### Cockpit is firewall friendly
+People using Cockpit should feel they are interacting with the underlying Server OS. Cockpit can look different on different operating systems, because it’s the UI for the OS, and not a external tool.
```
martinpitt: a"n" external tool

ideals.md
martinpitt: Also, this isn't actually what happens - we don't "remove" branding during  packaging, we never have it in the first place. I would rather say this directly, like "... not an external tool. Thus the UI should never actually show the 'Cockpit' name, but call the server by its name or operating system." (The exception is the "About.." menu entry, not sure if we really need to spell that out - I think we shouldn't in the interest of brevity.)

ideals.md
```diff
@@ -10,75 +10,52 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
 
-Cockpit opens one port for browser connections: by default that is 9090. We work well in cases where a bastion host runs Cockpit and connects out to the other servers via SSH.
+When packaging a stable release of Cockpit for release with a given OS, most of the “Cockpit” branding should be removed, and OS specific branding is put in its place.
 
-### Cockpit helps the server evolve coherently
+### Cockpit is zero footprint
 
-When developing Cockpit we run into situations where we notice the underlying server behavior is not coherent. We try to resolve that by contributing in the usual open source fashion, before implementing a hack or work around.
+The job of a server is to serve stuff to others. The tools to configure the server itself should not waste resources needed for that. As such, Cockpit has (as near as makes no difference) zero memory and process footprint on the server when not in use. Disk space usage of Cockpit is minimal and the required set of dependencies is scrutinized.
```
martinpitt: "serve stuff to others" sounds sloppy and does not really explain anything IMHO. I think that first sentence should just be dropped without replacement.

ideals.md
martinpitt: "try to as little API" → this sentence no verb. "have" or "expose"? The rest of the sentence sounds a bit  cumbersome too, so how about "…different release cadences. To achieve that, keep the `cockpit.js` Javascript API minimal, stable, and backwards compatible, and expose no other APIs".

ideals.md
larskarlitski: Thanks, good point! I didn't catch that (the diff shows this as new, but I've only moved it around). I've used @martinpitt's suggested sentence here. I think saying "but call the server by its name or operating system" implies that we get this information from the server. I also don't want to get too technical and detailed.

ideals.md
larskarlitski: Indeed. Thanks!

ideals.md
larskarlitski: I agree. I was unsure whether to leave it in or take it out. Taking it out now that you think that's preferable, too.

larskarlitski:
Updated with your suggestions. Thanks a lot for the feedback!

ideals.md
```diff
@@ -10,75 +10,54 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
 ### Cockpit is discoverable
 
-It should be possible to discover how to perform tasks through Cockpit without reading a help file. Astute readers may note that we have not completely acheived this yet, but this drives design decisions for Cockpit.
+It should be possible to discover how to perform tasks through Cockpit without reading a help file.
 
-### Cockpit is remotable
+### Cockpit directly interacts with and reacts to the system
 
-Cockpit can be effectively used when you are not sitting next to the server. This may seem obvious. Well, it is obvious.
+Cockpit does not store its own opinion of the state of the server anywhere. The UI always reflect the actual current state of the server. Cockpit reacts to system changes immediately, using the same system APIs that are used by command line tools.
```
martinpitt: reflect"s"

ideals.md
martinpitt: Thanks. The current version doesn't include the "keep the API minimal" aspect, but it's sort of implied.

ideals.md
```diff
@@ -10,75 +10,54 @@ These ideals also help define the scope of Cockpit. That scope is limited, and n
...
+### Cockpit is only the project name
 
-### Cockpit is firewall friendly
+People using Cockpit should feel they are interacting with the underlying Server OS. Cockpit can look different on different operating systems, because it’s the UI for the OS, and not an external tool.
 
-Cockpit opens one port for browser connections: by default that is 9090. We work well in cases where a bastion host runs Cockpit and connects out to the other servers via SSH.
+Thus the UI should never actually show the 'Cockpit' name, but call the server by its name or operating system.
 
-### Cockpit helps the server evolve coherently
+### Cockpit is zero footprint
```
martinpitt: IMHO Cockpit "has" a footprint, it "is" not one.

larskarlitski:
Thanks for catching those @martinpitt! Fixed both issues.

PR #149, opened 2018-03-19
Link to Contributing wiki page instead of duplicating it

martinpitt:
The page is out of date and there is no reason to have two versions of
it. Just link straight to the wiki.


martinpitt:
Try this out here: https://martinpitt.github.io/cockpit-project.github.io 

The "Contribute" (header) and "Contributing" (footer) links now point straight to the wiki.

garrett:
This is a huge step backward.

Main site navigation should not link offsite. It makes for a jarring experience. The page linked from GitHub didn't change for a few years when I had it imported into the site.

It's easy to update the extenals, just run the script to update:

```
_scripts/update-external-docs.rb
```

(It's in a `_scripts` directory, as anything with `_` in front is automatically not included in the output by Jekyll when building the site.)

garrett:
The script is currently ran manually, but was made with the idea of having a bot do it.

martinpitt:
Well, I think it's a step backwards (and very confusing) to have two non-identical copies of a page, but *shrug* - let's go with #150 then.

garrett:
Ideally, it shouldn't be non-identical; the script should auto-import it — but it has to be ran.

If it's on the website, it has an integrated look and feel and is also searchable. If it isn't, then it's not searchable and is a jarring experience when navigating.

PR #180, opened 2018-06-07
scale videos better when page width is small

garrett:


garrett:
Preview @ https://garrett.github.io/cockpit-project.github.io/blog/cockpit-169.html (resize browser width to be smaller)

PR #199, opened 2018-09-03
bump up supported browser versions (& remove IE)

garrett:
Also dropped the +, as the header is "minimum... versions"

### DIFF
--- a/_data/browsers.yml
+++ b/_data/browsers.yml
@@ -1,28 +1,24 @@
 - firefox:
   vendor: Mozilla
   name: Firefox
-  version: 11+
+  version: 23
 
 - chrome:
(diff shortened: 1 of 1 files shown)


garrett:
Similar to how our revised login page works (https://github.com/cockpit-project/cockpit/pull/9780), numbers were determined based on support for:

- websocket
- JSON
- CSS `@supports`
- JavaScript supports API

...As mentioned on caniuse.com.

In every case, the JS supports API was the highest version, and the new version numbers are reflected in the file.

garrett:
Changing the data updates the running page. Here's a cropped screenshot:

screenshot_2018-09-03 running cockpit cockpit project

PR #218, opened 2018-10-24
add contributor code of conduct v.1.4.1

garrett:


garrett:
TODO:

- [ ] Add contact email address
- [ ] Link to Code of Conduct from the front page footer
- [ ] Link to Code of Conduct from the Cockpit repo's README.md (this should be done after the PR is merged)

code-of-conduct.md
```diff
@@ -0,0 +1,80 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the project team at [INSERT EMAIL ADDRESS]. All
```
garrett: > [INSERT EMAIL ADDRESS]

We need to add a contact email address before merging the PR.

garrett:
Preview: https://garrett.github.io/cockpit-project.github.io/code-of-conduct.html

code-of-conduct.md
larskarlitski: There's no non-public email address for the full team and I'd prefer to be able to contact people personally. I'm fine with being one of the contacts (lars@karlitski.net). I think we should have more than one, though.

martinpitt: I'm fine with becoming a point of contact, too. In the interest of some more diversity here I would prefer Sanne or Jenn, though.

martinpitt:
Thanks, this looks reasonable to me. It expresses the intention clearly enough, but leaves the final decision what to do about mis-conducts to the project maintainers instead of prescribing a particular course of action; this is the right thing to do IMHO. And in fact it might even help *us*,  should it ever come to a similar situation as the recent [Opal indicent](https://www.reddit.com/r/codeofconduct/comments/3faovc/noncontibutors_force_opal_project_on_github_to/).

code-of-conduct.md
Gundersanne: I'll gladly be a point of contact, sanne.raymaekers@gmail.com

garrett:
Added contacts in alphabetical order (Lars, Martin, Sanne) @ https://github.com/cockpit-project/cockpit-project.github.io/pull/218/commits/852bfe9bf7aa7e6a107b6ba304de04c3315d022b — please know if you think there should be any more modifications, of if it looks good to you!

garrett:
Oh, I only added first names. Is that fine, or should I add last names too?

garrett:
I'll look into some simple obfuscation. Perhaps something a human can correct and some custom JS to auto-correct it, so if someone views the page with JavaScript on, they get it fixed after clicking the link and/or after a short timer or whatever.

PR #221, opened 2018-11-13
browser support: grid

garrett:
Updated minimum browser support requirements to require CSS grid.

### DIFF
--- a/_data/browsers.yml
+++ b/_data/browsers.yml
@@ -1,24 +1,24 @@
 - firefox:
   vendor: Mozilla
   name: Firefox
-  version: 23
+  version: 52
 
 - chrome:
   vendor: Google
   name: Chrome
-  version: 28
+  version: 57
 
 - edge:
   vendor: Microsoft
   name: Edge
-  version: 12
+  version: 16
 
 - ios:
   vendor: Apple
   os: iOS
   name: Safari
-  version: 9
+  version: 10.3
 
 - opera:
   name: Opera
-  version: 21
+  version: 44


andreasn:
I think this makes sense. I also wonder if what we do automated testing on should influence the browser support. We test on Firefox 63 and not 52 for example.

stefwalter:
This doesn't catch problems where we connect to an older version of cockpit from a newer login page or vs. versa. 

 * The login page my erroneously complain that required support is not available, but pages loaded from another system do not actually require grid.
 * The login page my not complain, but pages loaded from another system may require grid support.

I don't have a good answer to this, but it's something to be aware of.

martinpitt:
@stefwalter 's comment was probably meant to go to https://github.com/cockpit-project/cockpit/pull/10569

Anyway, let's go ahead with that, as we also want to land the above. In the worst case our web page is a little stricter than necessary, which doesn't hurt.

garrett:
Worst-case scenario when a browser that doesn't support grid is fed CSS that references grid:

- It uses an alternately specified fallback layout method (flex, float, block, inline), if one exists.
- If flex, float, block, or inline are not previously explicitly mentioned anywhere, then elements will be laid out according to standard block or inline properties inherent to the HTML element.

In other words: The extreme-corner-case browser would still see and be able to interact with everything. It wouldn't be broken at all; it would just do a standard HTML flow layout instead of a more advanced layout (if no alternate layout is already in place).

The odds of this happening are approximately ~1% of browsers that might happen to sign in using a Cockpit login instance that doesn't restrict non-grid browsers, connecting to a machine that uses some grid CSS for layout that doesn't have a fallback. It would be extremely unlikely, if it would ever happen.

Note: As far as I'm aware, we currently do not have any specified grids that do not have a fallback, except for mobile mode (which is only triggered when the browser is very narrow).

PR #226, opened 2018-12-06
Fixed typo in directory name.

rmcaninch:
Looked like a typo.


martinpitt:
Well spotted! However, this part is just autogenerated. It needs to be fixed here: https://github.com/cockpit-project/cockpit/blob/master/doc/guide/feature-systemd.xml

Do you want to send a PR? Thanks!

rmcaninch:
Sent a different PR. https://github.com/cockpit-project/cockpit/pull/10779

PR #227, opened 2018-12-12
Cockpit 184

mvollmer:
https://mvollmer.github.io/cockpit-project.github.io/blog/cockpit-184.html


mvollmer:
@garrett, would you like to add a paragraph about the new form layout?

mvollmer:
@martinpitt, would you like to add a paragraph about Cocktop on the Deskpit?

garrett:
Yep, good idea. I'll add in something. It's not too user-facing, but it is a little bit... and it's probably interesting for web developers too, as I haven't seen anyone use CSS grid for forms before.

garrett:
I don't think we should advertise "Cockpit on the Desktop". The primary goals of desktop integration are:

1. **Developer usage**
   - This is as part of a development workstation, but with server software installed, for developing software.
2. **Administrator usage**
   - It's a bastion host use case, where someone runs Cockpit locally and connects to remote machines that only have SSH exposed, so they don't have to expose 9090 on any servers.
   - Right now, this can be done manually by running Cockpit locally, going to port 9090 on localhost, and accepting a cert (often on the first time, but some browsers require it every time).
   - We could, however, have a custom WebKit browser, much like we do now for system registration, but optimized for Cockpit usage. This may automatically handle cert issues. It would abstract away the URL. It would enforce a sandbox separate from a browser with add-ons that could do malicious things. There are good reasons why we could consider a custom lightweight browser (which would be optional, of course).

What's not included in "Cockpit on the Desktop":

3. **Workstation usage**. Cockpit is for servers, with user interfaces specifically tuned for server usage.

Stating "Cockpit on the Desktop" would be very, very misleading for people, as they'll immediately jump to # 3, which is completely counter to the whole point of a server interface like Cockpit.

---

So, then, what is this "desktop integration" that we now currently have? It's a custom browser to use a part of Cockpit to register a RHEL system. It's outside the scope of "Cockpit on the Desktop" even if it is a little desktoppy. The fact that it's Cockpit is a quick-fix to provide a launchable UI of some sort for registration on workstations. This could (and probably should) be handled by a native desktop app in the future.

martinpitt:
@garrett: Yes, I wouldn't advertise it as a direct feature for users, just as a way how developers can integrate a cockpit page on the desktop. I'll try to come up with something decent.

_posts/2018-12-12-cockpit-184.md
```diff
@@ -0,0 +1,31 @@
...
+Cockpit is the [modern Linux admin interface](https://cockpit-project.org/). We
+release regularly.  Here are the release notes from version 184.
+
+### Machines: Dialog and tab layout is now responsive
+
+### Storage: Filesystem labels are validated up front
+
+### Storage: Some mount options are prefilled when needed
+
+### Cockpit on your Desktop (preview)
```
martinpitt: How about that:

### Integration of Cockpit pages on the desktop

For some use cases it may be desirable to use a particular Cockpit page on a desktop, if a native application does not exist. This can be done with the new "cockpit-desktop" tool, which runs a Cockpit web server and a web browser (Webkit, Firefox, or Chromium/Chrome) in an isolated network namespace. This avoids having to enable Cockpit system-wide and having to log into it through the web browser again.

cockpit-desktop may be used by system integrators or local administrators through a standard [desktop launcher](https://standards.freedesktop.org/desktop-entry-spec/latest/). Cockpit ships no pre-defined launchers by itself.

_posts/2018-12-12-cockpit-184.md
```diff
@@ -0,0 +1,60 @@
...
+summary: Cockpit with machines and storage improvements
+comments: 'true'
+---
+
+Cockpit is the [modern Linux admin interface](https://cockpit-project.org/). We
+release regularly.  Here are the release notes from version 184.
+
+### Machines: Dialog and tab layout is now responsive
+
+### Storage: Filesystem labels are validated up front
```
martinpitt: According to dictionary it's "up-front" or "upfront".

_posts/2018-12-12-cockpit-184.md
```diff
@@ -0,0 +1,60 @@
...
+launcher](https://standards.freedesktop.org/desktop-entry-spec/latest/). Cockpit
+ships no pre-defined launchers by itself.
+
+### Try it out
+
+Cockpit 184 is available now:
+
+ * [For your Linux system](https://cockpit-project.org/running.html)
+ * [Source Tarball](https://github.com/cockpit-project/cockpit/releases/tag/184)
+ * [Fedora 28](https://bodhi.fedoraproject.org/updates/cockpit-184-1.fc28)
```
martinpitt: We stop doing F28 releases now.

_posts/2018-12-12-cockpit-184.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Storage: Filesystem labels are validated up front
+
+The names one can give to filesystems have some limits put on them,
+and Cockpit now knows about them and tells you if you violate them.
+
+![Filesystem label validation](/images/storage-label-validation.png)
+
+This prevents some garish error messages from the underlying tools (I
+am looking at you, mkfs.xfs), and it also prevents silent truncation.
```
martinpitt: Let's avoid the snide comment in parentheses here :)

_posts/2018-12-12-cockpit-184.md
```diff
@@ -0,0 +1,60 @@
...
+slug: cockpit-184
+category: release
+summary: Cockpit with machines and storage improvements
+comments: 'true'
+---
+
+Cockpit is the [modern Linux admin interface](https://cockpit-project.org/). We
+release regularly.  Here are the release notes from version 184.
+
+### Machines: Dialog and tab layout is now responsive
```
martinpitt: How about something like "These UI elements now adjust their layout to small and mobile browser windows", and using a screenshot like

machines-create-vm-mobile

PR #240, opened 2019-04-17
Cockpit 192

martinpitt:
Rendered preview: https://martinpitt.github.io/cockpit-project.github.io/blog/cockpit-192.html


_posts/2019-04-17-cockpit-192.md
```diff
@@ -0,0 +1,56 @@
...
+set automatically after specifying the installation sources. This uses the
+libvirt [osinfo-detect](https://libosinfo.org/) utility.
+
+![Machines OS auto-detection](/images/machines-os-autodetect.png)
+
+### Translation cleanup
+
+The language dialog now offers all available translations instead of just a
+fixed subset, orders them alphabetically, and all languages are now shown with
+their native translation:
```
garrett: This paragraph is a little odd, as it goes from now to not to features of now.

I'd suggest something like this:

> Instead of showing just a predetermined set of languages, the language dialog now offers all available translations, ordered alphabetically, in their native translations.

_posts/2019-04-17-cockpit-192.md
```diff
@@ -0,0 +1,56 @@
...
+![New language dialog](/images/language-dialog-translated.png)
+
+Also, translations with less than 50% completion coverage got removed.
+
+### Allow accounts with non-standard shells
+
+Previously, logging into Cockpit was restricted to users who have a shell
+specified in [/etc/shells](https://linux.die.net/man/5/shells), enforced
+through [`pam_shells`](https://linux.die.net/man/8/pam_shells). This was found
+to be too strict (desktop logins don't make this restriction), and excludes
```
garrett: > (desktop logins don't make this restriction)

Not sure if this is relevant?

_posts/2019-04-17-cockpit-192.md
```diff
@@ -0,0 +1,56 @@
...
+users with e. g. the `tlog-rec-session` shell for
+[session recording](https://github.com/Scribery/tlog).
+
+Now this kind of shells that allow the user to login, but not change by
+themselves, are accepted for Cockpit login now. Shells like /sbin/nologin or
+/bin/false, which don't allow a user to login, continue to be denied for
+Cockpit as well.
```
garrett: This paragraph is very clunky.

Perhaps something like:

> Cockpit now accepts shells that are unchangeable by non-administrastors, yet still allow a user to log in. However, shells which do not allow a user to log in, such as `/sbin/nologin` or `/bin/false`, may still be used to deny access to Cockpit.

_posts/2019-04-17-cockpit-192.md
```diff
@@ -0,0 +1,56 @@
...
+---
+
+Cockpit is the [modern Linux admin interface](https://cockpit-project.org/). We
+release regularly.  Here are the release notes from version 192.
+
+### Machines: Auto-detect guest operating system
+
+When creating a new VM, the "OS Vendor" and "Operating System" fields are now
+set automatically after specifying the installation sources. This uses the
+libvirt [osinfo-detect](https://libosinfo.org/) utility.
```
garrett: I think @KKoukiou mentioned this detection only works for a small subset of Linuxes — specifically RPM-based distributions. (And even then, possibly not _all_ RPM-based distros.)

Is this correct? Should we state something about this limitation?

It seems like a pretty big oversight if we cannot detect Debian, Ubuntu, Arch, Alpine, BSDs, and _especially_ Windows.

_posts/2019-04-17-cockpit-192.md
martinpitt: It's a comparison how Cockpit is more akin to a desktop (GNOME/KDE) login than an ssh login. But I'm happy to drop the parentheses, they are just some more justification/explanation, but maybe too technical.

_posts/2019-04-17-cockpit-192.md
martinpitt: Seems to be quite comprehensive:
```
❱❱❱ osinfo-detect http://releases.ubuntu.com/18.04.2/ubuntu-18.04.2-live-server-amd64.iso
Media is bootable.
Media is an installer for OS 'Ubuntu 18.04 LTS'
Media is live media for OS 'Ubuntu 18.04 LTS'

❱❱❱ osinfo-detect https://cdimage.debian.org/debian-cd/current/i386/iso-cd/debian-9.8.0-i386-netinst.iso
Media is bootable.
Media is an installer for OS 'Debian 9'
```

But not perfect indeed:
```
❱❱❱ osinfo-detect http://dl-cdn.alpinelinux.org/alpine/v3.9/releases/x86_64/alpine-standard-3.9.3-x86_64.isoMedia is bootable.
```
I'll reword it to mention "attempt".

martinpitt:
@garrett: Many thanks for your excellent suggestions and the optimization! Pushed.

garrett:
As usual, I optimized the screenshots with `pngquant` and `oxipng` (active Rust fork of optipng):

Here's my `pngquant` alias (I have it aliased to `pq`):
```sh
pngquant --nofs -f --ext .png --skip-if-larger -v
```

This quantizes the image to an optimal pallete without dithering and saves it to the same file, overwriting the original (which is fine to do with a copy or in a git repo). If it doesn't find a solution to make the file smaller, it keeps the original. Its normally defaults to dither and saving as a different filename, hence the alias.

And here's what I do for oxipng:
```sh
oxipng -o6
```

Both commands are followed by the filename and only replace the file when it's smaller than the original. `oxipng` is quite safe to use, but `pngquant` is definitely lossy (although often without visual differences). Therefore, I _always_ use `oxipng` and also _try_ to use `pngquant`, unless the image looks obviously different upon inspection.

In these two images, `pngquant` worked well.

A lot of our screenshots can be compressed quite a bit.

It's nice to shave off image sizes when possible; it makes the blog posts load more quickly and it means it takes up less space in the git repo and on our hard drives.

_posts/2019-04-17-cockpit-192.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Translation cleanup
+
+Instead of showing just a predetermined set of languages, the language dialog
+now offers all available translations, ordered alphabetically, in their native
+translations:
+
+![New language dialog](/images/language-dialog-translated.png)
+
+Also, translations with less than 50% completion coverage got removed.
```
garrett:
```suggestion
Also, translations with less than 50% completion coverage were removed.
```

martinpitt:
Thanks @garrett! Applied, this can get squashed on merging.

PR #248, opened 2019-05-15
Updated running.md / fixed Debian 8 repo

MelonSmasher:
Updated the Debian 8 repo. The jessie-backports-sloppy url has been moved to archive.debian.org.


martinpitt:
As Jessie is EOL, let's please rather drop Debian 8 instructions entirely. Thanks for pointing out! I sent PR #249 about that.

PR #249, opened 2019-05-27
running: Drop obsolete Debian 8 instructions

martinpitt:
Debian 8 (Jessie) has been EOL for a while, and its backports repository
moved to the Archive.


martinpitt:
Rendered: https://martinpitt.github.io/cockpit-project.github.io/running.html#debian

Issue #269, opened 2019-08-14
Browser version numbers are misleading


garrett:
Using browser versions on our running/install page (https://cockpit-project.org/running.html) is misleading:

- There are numerous security issues in older browsers
  - Some of these issues are quite bad and could take over a system
  - Exploits exist in the wild and can attack some browser versions even just one release back (in some cases)
  - A compromised system can affect not just the system the browser runs on, but possibly any system it connects (using any connection method, like SSH, VNC, RDP, and even Cockpit)
   - Compromised systems can possibly run keyloggers (sniffing out passwords)
- It's tied to version numbers, whereas Cockpit checks for browser functionality, not specific browsers or versions (as it should).

For the reasons listed above, I suggest we refocus the browser version block on the install page and stress the importance of using up-to-date browsers instead of specific versions.

Perhaps we should even mirror Cockpit's login check with JavaScript on that page and show if the current browser is supported or not.

garrett:
For reference, here are the checks we perform on the login page:

https://github.com/cockpit-project/cockpit/blob/master/src/ws/login.js#L119-L176

PR #270, opened 2019-08-14
running: Port Cockpit's browser pass/fail tests

garrett:
Work-in-progress attempt to solve #269.

Right now, I have ported over Cockpit's browser feature detection to the website.

The intent is:
- show a simple pass/fail for the current browser
- display the reason why the browser does not work
- _easily_ copy/paste the code whenever we change the tests


garrett:
## TODO

- [ ] Display results in-page
- [ ] De-emphasize browser versions

garrett:
I'll be on vacation as of this evening, so I simply wanted to share the code I have, with the intent to finish this PR and merge sometime after August.

Right now, the code is loaded in our common JavaScript file, but only active on the running page. It displays the current result in the browser's console. This isn't good enough for production and would need to be extended to show something in page (hence the TODOs). But it should be enough to see that it works as-is right now. :wink:

garrett:
Updated branch with:
- Cockpit's login.js tests cross-referenced with caniuse.com's database to produce accurate versions
- Demo of current browser test based on the above tests (needs work, but there's something visible now)

garrett:
Success:

Screenshot_2019-10-23 Running Cockpit — Cockpit Project

Fail (simulated):

Screenshot_2019-10-23 Running Cockpit — Cockpit Project(1)

garrett:
I currently have this branch pushed to my master fork @ https://garrett.github.io/cockpit-project.github.io/running.html

_data/browsers.yml
```diff
@@ -1,24 +1,20 @@
...
   vendor: Microsoft
   name: Edge
-  version: 16
 
-- ios:
+- safari:
   vendor: Apple
   os: iOS
   name: Safari
   version: 10.3
```
martinpitt: Did you keep that version intentionally, or just forgot to remove it?

garrett: No, it's intentional. It's a manual minimum version. If it's higher than the computed one, it sticks. It's this version because of the certificate issues that exist on macOS Safari < 10.3.

(I also moved from referring to iOS for Safari to just Safari.)

garrett:
I need to squash this down. :wink:

_data/browsers.yml
garrett: Here's the bit of comparison code to support manual versions:

https://github.com/cockpit-project/cockpit-project.github.io/pull/270/files#diff-4283f68ad6ffc7bb5e893943e7f1b7dfR55-R67

It's a little bit of a mess, but that's because I'm doing this in liquid (which is the right place for this, in this case). I basically just compare the two values as floats.

martinpitt: Makes sense, thanks! Can you please add a comment about it, so that we can tell our future selves if/when to bump it?

garrett: It's a separate commit of the 3 and now has a bunch of text in the commit to explain it better.

garrett:
I have promise finally in the script, but commented out for now. It works and we can remove the comment and update the data when that lands. (It will land at some point in the future, when Edge is yet-another-Chromium-browser and Edge 18 is not-the-current-Edge.)

PR #275, opened 2019-09-30
Removed unnecessary padding blocking click in Guide, class .chapters

iokaravas:
I noticed an issue with some elements 'overflowing' their padding property, with no apparent visual result that blocked clicking links. Issue was most apparent in the Features section. Tested this in Chrome and Firefox (both on Windows).

The fix involved removing padding from said element's class. There is no apparent visual changes in the page or in any other pages (classification was under the page so I didn't expect that).

**Before:**
<kbd>
unclickable
</kbd>

**After:**
<kbd>
clickable
</kbd>

Also removed a stray comma.

### DIFF
--- a/assets/site/guide.sass
+++ b/assets/site/guide.sass
@@ -141,8 +141,7 @@ html.cockpit-guide
       text-decoration: underline
 
   h2, h3, h4, h5, h6,
-  dl.toc .part,
-  dl.toc .chapter,
+  dl.toc .part
     position: static
     padding: 1rem 0 0.5rem
 
--- a/assets/site/guide.sass
+++ b/assets/site/guide.sass
@@ -149,7 +149,7 @@ html.cockpit-guide
   dl.toc .part
     font-size: 1.72rem
 
-  h3,
+  h3
     font-size: 1.4rem
 
   .refnamediv h2,


assets/site/guide.sass
```diff
@@ -142,15 +142,13 @@ html.cockpit-guide
 
   h2, h3, h4, h5, h6,
   dl.toc .part,
-  dl.toc .chapter,
```
KKoukiou: This is the change that you describe in the commit message I guess.

assets/site/guide.sass
```diff
@@ -142,15 +142,13 @@ html.cockpit-guide
...
     position: static
     padding: 1rem 0 0.5rem
 
-  h2,
```
KKoukiou: But this change here does not correspond to the change mentioned in the commit message.

assets/site/guide.sass
```diff
@@ -142,15 +142,13 @@ html.cockpit-guide
...
-  dl.toc .part
-    font-size: 1.72rem
+  dl.toc .chapter
+    position: static
 
-  h3,
+  h3
```
KKoukiou: Neither this.

assets/site/guide.sass
iokaravas: Yes this the one.

assets/site/guide.sass
```diff
@@ -142,15 +142,13 @@ html.cockpit-guide
 
   h2, h3, h4, h5, h6,
   dl.toc .part,
-  dl.toc .chapter,
     position: static
     padding: 1rem 0 0.5rem
 
-  h2,
-  dl.toc .part
-    font-size: 1.72rem
```
iokaravas: These two should stay, that's an error from my part.

assets/site/guide.sass
```diff
@@ -142,15 +142,13 @@ html.cockpit-guide
...
+  dl.toc .chapter
+    position: static
```
iokaravas: This is to keep the same styling for the element as before but without the padding. So it is related to the commit message.

assets/site/guide.sass
iokaravas: I should have made a different commit from my understanding , but this is just a trailing comma, it's no error but it's unnecessary.

garrett:
@iokaravas: Thanks for your contribution! The CSS for the docs is a bit of a mess, as it's having to undo and redo a bunch of CSS that come from our docs generator.

I reviewed the code and rewrote the commits (attempting to preserve authorship, but it doesn't seem to have worked).

I've looked over the code and the resulting pages and it seems we don't need to set static on `dl.toc .chapter` either, so I simply removed it.

Preview @ https://garrett.github.io/cockpit-project.github.io/guide/latest/

assets/site/guide.sass
garrett: I moved this to a separate commit.

PR #285, opened 2019-11-13
index: Make badge take "and" into consideration

garrett:
Assumption: Cockpit will always be the first release in a title.


garrett:
This is extremely naïve code, but it's in Liquid, so that's what we get. :wink:

PR #287, opened 2019-11-27
Document support for Fedora CoreOS

martinpitt:
Preview: https://martinpitt.github.io/cockpit-project.github.io/running.html


running.md
```diff
@@ -134,6 +134,21 @@ sudo firewall-cmd --add-service=cockpit
 sudo firewall-cmd --add-service=cockpit --permanent
 ```
 
+### Fedora CoreOS
+{:#coreos}
+
+The standard Fedora CoreOS image does nto contain any Cockpit packages. To install these, run
```
davidmalcolm: typo: "nto";  should presumably be "not"

martinpitt: Thanks David, well spotted! Fixed.

martinpitt:
As per IRC review, this is blocked on configuring the container to auto-start on boot, with `podman generate systemd`. For that we should support setting a fixed name, and possibly wrap that generation, daemon-reload and unit enable into another label. So parking this as a separate PR for now.

PR #288, opened 2019-11-27
Document support for Fedora CoreOS

martinpitt:
As per IRC review, this is blocked on configuring the container to auto-start on boot, with podman generate systemd. For that we should support setting a fixed name, and possibly wrap that generation, daemon-reload and unit enable into another label. So parking this as a separate PR for now.

 - [x] https://github.com/cockpit-project/cockpit/pull/13210


martinpitt:
For cockpit release notes:
```
### Support Fedora CoreOS

[Fedora CoreOS](https://docs.fedoraproject.org/en-US/fedora-coreos/) is a minimal and robust, Fedora based operating system for hosting container workloads. It is the successor of [Fedora Atomic Host](http://www.projectatomic.io/), which is not supported any more.

Please see the [Running Cockpit](https://cockpit-project.org/running.html) documentation or the [cockpit/ws docker hub](https://hub.docker.com/r/cockpit/ws/) page for how to install Cockpit on Fedora CoreOS.

```

garrett:
There's a FIXME in the CoreOS instructions for the command to make Cockpit start on boot.

martinpitt:
https://github.com/cockpit-project/cockpit/pull/13210 landed now, I synced the changes to cockpit-container.git, https://hub.docker.com/r/cockpit/ws is now updated as well. I close this PR in favor of #290

PR #297, opened 2020-01-22
image: Add new Edge logo

garrett:
Moved old Edge logo to legacy and added new Edge logo.


garrett:
I couldn't find a vector version of the Edge logo, so I made one with a quick trace and radial gradients (instead of meshes), so it's approximate. The differences fall away when it's displayed at standard sizes on screens, however, so it looks right on our site. :wink:

garrett:
Screenshot:

Screenshot_2020-01-22 Running Cockpit — Cockpit Project

PR #298, opened 2020-01-23
Update external docs (2020-01-23)

garrett:
Update to the latest versions of HACKING and README


garrett:
The Ruby script depends on Ruby, therefore, I ran it in the website container like thus:

```sh
toolbox run -c cockpit-website ./_scripts/update-external-docs.rb
```

PR #302, opened 2020-02-20
Added a WIP script to update the docs with toolbox

garrett:
I started to write a script to automate the doc updating process.

It kind of works, but needs to be finished. I ran out of time. The if/then isn't working right now, and I don't know why.

I guess you can see my intent?

`git -d` won't delete a branch unless it has already been merged or is the same as master, so it should delete a branch if there are no changes... and it won't (with an error code) if the branch hasn't been merged yet.

### DIFF
--- a/_scripts/toolbox-doc-update
+++ b/_scripts/toolbox-doc-update
@@ -1,0 +1,23 @@
+#!/bin/bash -e
+
+date=$(date -I)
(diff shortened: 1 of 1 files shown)


martinpitt:
I propose for now to just let that run fully automatically, see PR #364.

martinpitt:
Superseded by #364

PR #310, opened 2020-03-23
running: Update Arch installation instructions

martinpitt:
Show the commands how to install and enable it.

The Community Repository is part of the official repositories, so tone
down the text to stop making it sound exceptional/unsupported.

Also mark cockpit as "included" for the same reason.

Preview: https://martinpitt.github.io/cockpit-project.github.io/running.html

### DIFF
--- a/_data/distros.yml
+++ b/_data/distros.yml
@@ -46,3 +46,4 @@
 
 - archlinux:
   name: Arch Linux
+  included: true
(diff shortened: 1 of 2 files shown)


garrett:
We might want to add a one-line instruction to visit IP address at 9090 in their browser and a link to more information. Just to improve flow. But this would be for every set of instructions and a different PR.

PR #314, opened 2020-04-19
running: Avoid partial upgrade in Arch installation instructions

### LINKED ISSUE #313: Unsupported options on Arch instructions
Hello,
The instructions for [installing Cockpit on Arch](https://cockpit-project.org/running.html#archlinux) currently uses unsupported options for `pacman`:
```
sudo pacman -Sy cockpit
sudo systemctl enable --now cockpit.socket
```

Using `pacman -Sy` causes a partial upgrade, which is [unsupported](https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported) and could potentially cause breakage (the new package will pull any updated libraries, which may break packages that haven't been updated yet).

I recommend updating the instructions to:
```
sudo pacman -Syu && sudo pacman -S cockpit
sudo systemctl enable --now cockpit.socket
```

This will perform a full upgrade (`pacman -Syu`) first, then install Cockpit (`pacman -S cockpit`).

martinpitt:
`pacman -Sy <package>` is an unsupported operation that should be
avoided [1] as it can cause a partial upgrade of the system. Instead,
just document how to install only the cockpit package, and separately
document the (uncommon) case when repository indexes don't exist.

Thanks to NEXCydran for pointing this out!

[1] https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported

Fixes commit 36605790126
Fixes #313

### DIFF
--- a/running.md
+++ b/running.md
@@ -242,8 +242,8 @@ sudo systemctl enable --now cockpit.socket
 
 [Cockpit](https://www.archlinux.org/packages/community/x86_64/cockpit/) is included in [Arch Linux](https://www.archlinux.org/packages/):
 ```
-sudo pacman -Sy cockpit
+sudo pacman -S cockpit
 sudo systemctl enable --now cockpit.socket
 ```
 
-
+If the first command fails with "database file for ... does not exist", refresh/update your system with `sudo pacman -Syu` first.


martinpitt:
Preview: https://martinpitt.github.io/cockpit-project.github.io/running.html#archlinux

@NEXCydran: Want to have a look?

NEXCydran:
Just took a look at the preview - looks good!

Thanks for updating the site, I believe that the Arch installation directions are accurate now.

PR #320, opened 2020-05-05
Add support for alerts, notes, warnings, etc.

### LINKED ISSUE #319: style: Add classes for notable blocks (warning, error, info, etc.)
The classes should be used like this within Markdown:

```md
_**Note**: This is an informative note!_
{:.info}

_**Warning**: This is a warning!_
{:.warning}

_**Danger**: This is a dangerous!_
{:.danger}
```

Both bolding and italics are used in the example to show things are important and allow screen readers and RSS feeds to show intent. The classes may override some of the visual style (perhaps the italics), but it would be a good idea to use both even with a class.

PF4 uses the following for alerts:

- alert (without specifying what type)
- info
- success
- warning
- danger

We should name the website similarly, so there's less that one needs to remember.

Of these, the ones that make the most sense to implement for the website are:

- info
- warning
- danger

It _might_ make sense to have an alias for some of these... most _notable_ would be to alias info to note, so both work as expected.

We could even consider basically re-using PF4-like styling of the inline alerts  @ https://www.patternfly.org/v4/documentation/react/components/alert#inline-types

garrett:
Fixes #319.


garrett:
BTW: The two commits are separate on purpose. One adds support, the other uses it across old files.

martinpitt:
This looks promising, thanks! As this is not from your master branch, is there an auto-generated preview somewher? Also, can you please add a `Closes #`?

garrett:
I've added icons and also have all the alert styles.

Preview: https://garrett.github.io/cockpit-project.github.io/blog/creating-plugins-for-the-cockpit-user-interface.html

garrett:
Checked in Firefox, Chrome, WebKit. It uses `@supports` to progressively add an icon. (Fallback is to not show an icon or even space for it if `mask: url(...)` isn't supported. It is supported in all our target browsers, however, so we're good.)

PR #331, opened 2020-06-30
Upgrade Gems

garrett:
Updated to the latest Gems.


garrett:
I had some issues building the F32 based toolbox (using the scripts in the repo) from scratch yesterday. Upgrading the gems helps. It seems Fedora 32 updated some packages which caused compilation errors with our old gems, but works fine with the new ones.

It was past time to update our gems again — but not that it matters much, as we compile to static anyway.

PR #333, opened 2020-06-30
External docs update  (2020-06-30)

garrett:
1. Add support for the bots README file in our docs updater script.
2. Sync all external docs to the website, including the bots README.

### DIFF
--- a/_scripts/update-external-docs.rb
+++ b/_scripts/update-external-docs.rb
@@ -15,6 +15,11 @@ source_pages = [
   "test/README.md",
 ]
 
+# Pages to fetch from the Cockpit bots repo
+bots_pages = [
+  "README.md"
+]
+
 # URLs to rewrite
 @url_rewrites = {
   "https://github.com/cockpit-project/cockpit/blob/master/HACKING.md" =>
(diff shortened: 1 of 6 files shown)


_scripts/update-external-docs.rb
```diff
@@ -73,7 +80,7 @@ def fetch_and_add(pages, type=:source)
     file = "#{type}/#{page}"
     file += ".md" unless page.match(/\.md$/)
 
-    doc = open(url_raw).read
+    doc = URI.open(url_raw).read
```
garrett, line 83: I snuck in a little fix here too. Calling `uri-open` in Ruby with just "open" is deprecated and throws a warning. Hence, adding `URI.` to be explicit on which open we want to use.

external/source/test/README.md
```diff
@@ -111,79 +105,36 @@ You can set these environment variables to configure the test suite:
...
-and other scripts that work with test machine images.
-
-    "fedora-NN" -- The basic image for running the development version of Cockpit.
-                   This is the default.
-
-    "ipa"       -- A FreeIPA server.
-
-    "openshift" -- An Openshift Origin server.
+for each test. See the
+[bots documentation](https://github.com/cockpit-project/bots/blob/master/README.md)
```
garrett, line 113: Oh, I need to add a rewrite to keep the bots link on our site, instead of going to GitHub.

martinpitt: Why does that not work? The commit doesn't really explain why this now pulls in bots README. Just for consistently looking formatting?

garrett:
> The commit doesn't really explain why this now pulls in bots README. Just for consistently looking formatting?

We pull in external docs to the website so someone browsing it has a consistent experience and so they're indexed with the search. Having all of the important Cockpit information in one place is useful and makes it more discoverable.

PR #340, opened 2020-08-18
Update gems (2020-08-18)

garrett:
Fixes Kramdown security issue CVE-2020-14001 (Unintended read access and code execution)
Details: https://nvd.nist.gov/vuln/detail/CVE-2020-14001

Naturally, this doesn't affect the website directly; it only affects building and running locally.

It's still a good idea to fix, however.


garrett:
Additionally, it keeps our gems up-to-date, which is a good idea, especially as F33 is not too far off. (Once in a while older Ruby gems don't want to build properly on newer Fedoras. It's less of a problem, as we usually rely on GitHub to build things. When building locally, we're generally using toolbox, so we can control the version of Fedora it builds in anyway.)

martinpitt (approved):
Thanks!

PR #345, opened 2020-09-15
Script fixes

garrett:
Minor fixes brought over from Springboard (where I noticed some issues from backporting Cockpit yesterday).

### DIFF
--- a/_scripts/toolbox-create
+++ b/_scripts/toolbox-create
@@ -8,7 +8,8 @@ run="toolbox run -c cockpit-website"
 echo "## Installing RPM dependencies inside container..."
 $run sudo dnf install -y rubygem-bundler \
 	ruby-devel libffi-devel make gcc gcc-c++ redhat-rpm-config zlib-devel \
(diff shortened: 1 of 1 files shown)


_scripts/toolbox-create
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/bash -e
 
 echo "## Creating website container..."
-toolbox create -c cockpit-website
+toolbox create -c cockpit-website -r 32
```
martinpitt, line 4: What's the reason for this (the commit message and/or a comment should point that out)? It'll get out of date quickly enough. Does f33 not work?

garrett: F33 does not work currently. The idea is to bump it up on every stable release, so we have a known good platform instead of relying on the host system's version, which might be in flux (like it is currently).

It would be nice if we could just do `-r stable`, but I don't think that's a thing?

martinpitt (approved):
Ack, thanks!

PR #351, opened 2020-10-02
make index page of resources - need improvement

andaugust:
an amazing project sir but plz make index page of resources


martinpitt:
Shame on you for abusing the hacktoberfest...

PR #355, opened 2020-11-24
Update external docs (2020-11-24)

garrett:
Updated to the latest external docs with the following command:

```sh
toolbox run -c cockpit-website sh -c _scripts/update-external-docs.rb
```

(Which should work on every system when there's a cockpit-website container, made with `_scripts/toolbox-create`.)


martinpitt (approved):
Thanks!

PR #357, opened 2020-12-02
added script to allow install software inside the toolbox

silverblue-devops:
usage: toolbox-install neovim python3-neovim

```
#!/usr/bin/bash -e

echo "## installing $@"

run="toolbox run -c pet"

$run sudo dnf install "$@"

echo "## Done!"
echo "## Run _scripts/toolbox-run"


```

### DIFF
--- a/_scripts/toolbox-create
+++ b/_scripts/toolbox-create
@@ -8,7 +8,8 @@ run="toolbox run -c cockpit-website"
 echo "## Installing RPM dependencies inside container..."
 $run sudo dnf install -y rubygem-bundler \
 	ruby-devel libffi-devel make gcc gcc-c++ redhat-rpm-config zlib-devel \
-	libxml2-devel libxslt-devel tar nodejs
+	libxml2-devel libxslt-devel tar nodejs \
+	glibc-all-langpacks
 
 echo "## Setting local gem path"
 $run bundle config path .gem
--- a/_scripts/toolbox-create
+++ b/_scripts/toolbox-create
@@ -1,7 +1,7 @@
 #!/usr/bin/bash -e
 
 echo "## Creating website container..."
-toolbox create -c cockpit-website
+toolbox create -c cockpit-website -r 32
 
 run="toolbox run -c cockpit-website"
 


_scripts/toolbox-create
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/bash -e
 
 echo "## Creating website container..."
-toolbox create -c cockpit-website
+toolbox create -c cockpit-website -r 32
```
martinpitt, line 4: Can you please give a rationale for this in the commit message, and a comment? I.e. why does the current version not work?

garrett: The F32 version was when F33 was beta and not working due to a bug. (This was in September.) The idea is that we'd have a consistent repo used.

We can bump it to F33 now. Or just drop it.

garrett:
The idea of using the toolbox container script is to use toolbox to isolate Jekyll and the whole Ruby stack from the host. We could use Podman itself for this, but then we'd have to map a bunch of stuff that toolbox does itself.

I never run anything inside, except for Jekyll (and updating the docs, but that's a special case and I use `toolbox run -c cockpit-website sh -c _scripts/update-external-docs.rb`).

As toolbox lets you launch other toolbox containers from within a toolbox container, I have a development toolbox where I have things like vim/neovim where I use that for development purposes and then I run the toolbox create and run scripts from within my development toolbox (and it "just works").

_scripts/toolbox-create
garrett: BTW: I'm not sure why this old commit is showing up on this PR. Weird.

garrett: Also, the reason why I specified the version is that the default toolbox version you'll have is depending on the version of Fedora you're running. If your host is Fedora 32, you'll have a Fedora 32 toolbox. If the host is Fedora 33, then you'll have a Fedora 33 toolbox. This was to make it consistent for everyone, regardless of the version of Fedora they are running. But as Fedora 33 is now stable and works, we can definitely bump up the version.

martinpitt: I thought that the contents of the toolbox isn't very sensitive to the underlying Fedora version, so hardcoding any version will just quickly get out of date and thus we should start with the default version until we run into trouble (at which point we can still pin it). But that's not a strong opinion, if you feel like updating it every now and then, 32 is ok as well.

_scripts/toolbox-create
```diff
@@ -8,7 +8,8 @@ run="toolbox run -c cockpit-website"
 echo "## Installing RPM dependencies inside container..."
 $run sudo dnf install -y rubygem-bundler \
 	ruby-devel libffi-devel make gcc gcc-c++ redhat-rpm-config zlib-devel \
-	libxml2-devel libxslt-devel tar nodejs
+	libxml2-devel libxslt-devel tar nodejs \
+	glibc-all-langpacks
```
martinpitt, line 12: You can get UTF-8 with setting `export LC_ALL=C.UTF-8` in the container, and save yourself 19 MB in the container.

martinpitt:
BTW, the title of this PR is really strange, considering what the commits actually do :grin:

garrett:
The commits are already in master and have been for several months. I don't understand what's going on and why they don't relate to what this PR is about.

_scripts/toolbox-create
garrett: Ah, good to know.

martinpitt:
Indeed, how very weird. Let's close this.

garrett:
JFWIW, if you do actually want to run stuff inside the Cockpit website toolbox container, it's easy with:

```sh
toolbox run -c cockpit-website COMMAND_HERE
```

And if you want to go inside and have it be interactive, there's:

```sh
toolbox enter cockpit-website
```

PR #362, opened 2020-12-15
scripts: Bump up toolbox-create to F33

garrett:
As we're controlling the version of Fedora in the toolbox, this makes sure anyone creating a new toolbox container is using Fedora 33, regardless of the version of Fedora used on their host. (Toolbox defaults to the same version as the host when creating new toolbox containers.)

As Fedora 33 has been stable for some time now, this is fine.


martinpitt (approved):
Thanks!

PR #366, opened 2021-01-04
include: Handle source override in GitHub edit

### LINKED ISSUE #235: Extend "edit this page" with frontmatter source
Externals have a source attribute in the frontmatter. We change the "edit this page" link to use that instead.

(In addition, we could also visibly mention somewhere that the document comes from a different place too.)

garrett:
Fixes #235

Includes support for both repo and wiki links. The link is to the edit mode on GitHub.

Our externals script has included "source:" with URLs in the frontmatter this whole time, with anticipation of this support. :wink:


martinpitt (approved):
Nice! Thanks!

PR #368, opened 2021-01-04
search: Use DuckDuckGo when JS is not enabled

### LINKED ISSUE #178: search: fall back to external search engine when JS is not enabled
The search engine should fall back to an external search engine when JavaScript is disabled. For the sake of no-tracking, I suggest DuckDuckGo.

garrett:
Fixes #178

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -7,8 +7,9 @@
 <div class="search-page">
 
   <!-- Search form -->
-  <form class="grid-noGutter search-form" id="search-form" role="form">
-    <input autocomplete="off" id="search-input" name="entry" class="col-10_xs-8 search-input" aria-labelledby="searchFormButton">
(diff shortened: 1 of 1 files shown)


martinpitt (commented):
How is that only used without js? Is there some `<noscript>` somewhere that doesn't appear in the diff?

garrett:
It's "progressive enhancement": It's a standard form that submits a query to DuckDuckGo. HTML forms work without any JavaScript. :wink:

The JS already has `preventDefault`, so it works as before when you have JavaScript enabled.

garrett:
Before this change, it didn't send the form to anywhere, so it posted it back on the same page. But as we have a static site, it didn't do anything.

I should've fixed this long ago... especially as it's so simple. Whoops. Better late than never.

martinpitt (approved):
Ack, thanks!

Issue #231, opened 2019-01-27
Fix horizontal scrollbars in the guide in Chrome


garrett:
In Chrome on Ubuntu in 1024x768, there's a horizontal scrollbar. It looks like content is flowing off the page?

garrett:
I believe I happened to fix this when adjusting the guide CSS a while ago.

PR #369, opened 2021-01-04
running: Adjust breakpoint for OS table

garrett:
I was going to address #231 but looked at the running page instead. I noticed it did have horizontal overflow due to the breakpoint being a little off. I've fixed it, even though it's unrelated to the original issue (which I think was fixed in another commit, a long while ago).


martinpitt (approved):
Danke!

PR #370, opened 2021-01-04
include/feed: Use absolute URL for images

garrett:
Some feed readers support referencing images relative to the `<link>` as `<base href>`,  others do not.

By changing our path referenced images to be absolute instead, _all_ feed readers will be supported (provided they stick to https to avoid mixed content, but that's on their side, not ours).

### DIFF
--- a/_includes/feed.html
+++ b/_includes/feed.html
@@ -4,23 +4,33 @@ assign posts = include.posts | default: site.posts
 assign limit = include.limit | default: site.paginate | default: 10
 %}{%
 assign title = include.title | default: site.title
+%}{%
+assign url_prefix = site.url | append: site.baseurl | append: '/'
 %}<?xml version="1.0" encoding="UTF-8"?>
 <rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
   <channel>
     <title>{{ title | xml_escape }}</title>
     <description>{{ site.description | xml_escape }}</description>
-    <link>{{ site.url }}{{ site.baseurl }}/</link>
-    <atom:link href="{{ "/feed.xml" | prepend: site.baseurl | prepend: site.url }}" rel="self" type="application/rss+xml"/>
+    <link>{{ url_prefix }}</link>
+    <atom:link href="{{ "/feed.xml" | prepend: url_prefix }}" rel="self" type="application/rss+xml"/>
     <pubDate>{{ site.time | date_to_rfc822 }}</pubDate>
     <lastBuildDate>{{ site.time | date_to_rfc822 }}</lastBuildDate>
     <generator>Jekyll v{{ jekyll.version }}</generator>
     {% for post in posts limit: limit %}
       <item>
         <title>{{ post.title | xml_escape }}</title>
-        <description>{{ post.content | markdownify | xml_escape }}</description>
+        <description>
+          {{ post.content
+            | markdownify
+            | replace: 'img src="/', 'img src="$SITE_GOES_HERE'
+            | replace: 'a href="/', 'a href="$SITE_GOES_HERE'
+            | replace: '$SITE_GOES_HERE', url_prefix
+            | xml_escape
+          }}
+        </description>
         <pubDate>{{ post.date | date_to_rfc822 }}</pubDate>
-        <link>{{ post.url | prepend: site.baseurl | prepend: site.url }}</link>
-        <guid isPermaLink="true">{{ post.url | prepend: site.baseurl | prepend: site.url }}</guid>
+        <link>{{ post.url | prepend: url_prefix }}</link>
+        <guid isPermaLink="true">{{ post.url | prepend: url_prefix }}</guid>
         {% for tag in post.tags %}
         <category>{{ tag | xml_escape }}</category>
         {% endfor %}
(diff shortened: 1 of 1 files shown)


_includes/feed.html
```diff
@@ -4,23 +4,32 @@
...
     <generator>Jekyll v{{ jekyll.version }}</generator>
     {% for post in posts limit: limit %}
       <item>
         <title>{{ post.title | xml_escape }}</title>
-        <description>{{ post.content | markdownify | xml_escape }}</description>
+        <description>
+          {{ post.content
+            | markdownify
+            | replace: 'img src="/', 'img src="$SITE_GOES_HERE'
+            | replace: '$SITE_GOES_HERE', url_prefix
```
garrett, line 26: BTW: Liquid (Jekyll's template language) is simple and a bit wonky, so you have to dance around with stuff like this. We want to match on the `img src="` and also keep it. I don't think it can get any more elegant, as it's just too simple.

garrett:
I tested it on https://garrett.github.io/cockpit-project.github.io/blog/feeds/all.atom.xml and it now works with images in NextCloud News.

_includes/feed.html
martinpitt: Ugh, what a hack -- string replacement in free-form fulltext HTML? Also, site-relative `` URLs are basic HTML, and in fact preferable to make downloading/mirroring actually work. I.e. if a feed reader actually downloads everything (which many Android apps do to make reading available offline), this *breaks* images for them.

martinpitt (commented):
IMHO this is a regression as it works around broken feed readers at the expense of breaking correct ones. I don't have a strong enough opinion to actively veto it, but I can't in good faith approve it. Anyone else of course feel free to :+1: if you think differently. Thanks!

garrett:
> IMHO this is a regression as it works around broken feed readers at the expense of breaking correct ones.

How does it break "correct" ones?

> string replacement in free-form fulltext HTML?

We have a lot of string replacement in HTML on the site. There's no other way. Liquid is limited. How else can this be done? If you want to look at the liquid docs and find some way this can be done better, please feel free. (You won't find one.)

Everything with `` block for some reason.)

> if a feed reader actually downloads everything (which many Android apps do to make reading available offline), this breaks images for them.

It sounds like _that_ feed reader is broken, if it downloads everything offline and doesn't rewrite it to work.

---

According to the validator, our Atom feed is valid, but the images should not be relative:
https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fcockpit-project.org%2Fblog%2Ffeeds%2Fall.atom.xml

Here's the info:

https://validator.w3.org/feed/docs/warning/ContainsRelRef.html

> Some RSS elements are allowed to contain HTML, including links and images. However, if these links use relative URLs to point to resources, users of some news aggregators will not be able to click on them. And your images use the relative URLs, the images may not appear (since the news aggregator will be looking for the images locally on the user's hard drive, rather than on your web site).

(We might need to fix relative links too. Fun.)

_includes/feed.html
garrett: Feel free to find a better solution. There isn't one. (This is Liquid, which is extremely limited.)

https://learn.cloudcannon.com/jekyll-cheat-sheet/
https://github.com/Shopify/liquid/wiki/Liquid-for-Designers

martinpitt: Right, I don't doubt that *if* you need to replace img src tags, then that is the simplest possible mechanics. Resolving this comment, as we have another one for that "if" question.

martinpitt:
> How does it break "correct" ones?

Suppose you download the HTML and images locally, with `wget -r` or a thing like planet, or a feed reader application. Then these hacked `<img src="https://cockpit-project/.../foo"` absolute URLs would download the images from our page, instead of the local copy?

garrett:
The literal RSS and Atom specs say to use absolute URLs for images and links. Anything that cannot cope with rewriting is doing it wrong.

If `wget` still uses the absolute URLs, then it was not invoked correctly. It should include `--convert-links` (or `-k` for short) in the arguments.

If you look at a planet, such as [Planet GNOME](https://planet.gnome.org/), [Fedora Planet](http://fedoraplanet.org/), or [Planet Debian](https://planet.debian.org/), you'll see that images on each are all absolute, hotlinking the original site.

garrett:
- [ ] We should make sure links are handled as well.

Also, RSS/Atom readers should handle both relative and absolute links correctly, regardless of spec. But I guess some don't, really. We can't fix that though (unless they're FOSS and we file upstream issues and/or send code changes their way).

martinpitt:
> The literal RSS and Atom specs say to use absolute URLs for images and links

Oh, that makes sense then, thanks! (It sounds rather weird intuitively, sorry)

martinpitt (approved):

garrett:
I've also:
- added support for rewriting links in the same manner (same commit)
- fixed the self link, as pointed out by the validator (separate commit, but still in this PR, as it's simple and related to the feed)

I'm not going to "fix" the validator's `<iframe>` warning, as that would remove support for inline videos, and there's nothing we can properly replace them with. Some feed readers will support them, others might not. 

(I suppose it might be possible to have a fallback bit of text and/or image indicating that the video isn't there.)

garrett:
Preview the changes in the validator @ https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fgarrett.github.io%2Fcockpit-project.github.io%2Fblog%2Ffeeds%2Fall.atom.xml

Everything's fixed, except the iframe. (It's a warning and I stated justification for why I'm not touching that, at least in this PR.)

martinpitt (approved):
Thanks! Approval still stands, please land at will.

PR #374, opened 2021-01-13
images: Convert CoreOS text to paths; optimize SVGs

garrett:
Someone else added these distro graphics to our repo. The CoreOS one had text still, so the text wasn't correct where people lacked the font. I converted it to paths. Meanwhile, both needed optimization in file size for the web as well (using [SVGO](https://github.com/svg/svgo) via [SVGOMG](https://jakearchibald.github.io/svgomg/)).


martinpitt (approved):
Danke!

PR #384, opened 2021-03-03
Update guide to version 239

allisonkarlitskaya:
Here's the commit that got missed because of https://github.com/cockpit-project/cockpit/pull/15461


martinpitt (approved):
Thanks for relieving the bots of some work *smirk*!

PR #387, opened 2021-03-11
running: Update browser support

garrett:
- Drop unused icons
- Optimize icons
- Split out second tier from first tier
- Add Safari & GNOME Web to second tier


garrett:
## Before

Screenshot_2021-03-11 Running Cockpit — Cockpit Project(1)

## After

Screenshot_2021-03-11 Running Cockpit — Cockpit Project(2)

garrett:
Edge no longer has automated tests, and we do care that WebKit still works, so we (probably especially I) check it.

martinpitt (approved):
Thank you for that! That's indeed more current and honest.

PR #393, opened 2021-04-05
post: Unified upstream and downstream testing with TMT and Packit

martinpitt:
Preview: https://martinpitt.github.io/cockpit-project.github.io/blog/fmf-unified-testing.html

I'd appreciate some reviews whether this is a coherent and comprehensive story, and of course also fixes to my broken English. Thank you!

@thrix, @TomasTomecek, @psss FYI

PR #394, opened 2021-04-06
Posts  headline summary and category

garrett:
Fixed summary mode to only use headlines; spotted in #393. Also added category with a link (when available).


martinpitt (approved):
Danke!

PR #395, opened 2021-04-07
style: Improve abbr display across browsers

garrett:
This makes `<abbr>` show up in all browsers with squiggles. Previously, WebKit had issues and didn't style it whatsoever.

### DIFF
--- a/assets/site.sass
+++ b/assets/site.sass
@@ -390,6 +390,18 @@ p.screenshot img
 
 abbr
   cursor: help
+  // Webkit does not support combining these values with a shorthand text-decoration
+  text-decoration: underline
+  text-decoration-color: #aaa
+  text-decoration-skip-ink: none
+  text-decoration-style: wavy
+  text-decoration-thickness: 1px
+  text-underline-offset: 0.2em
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks! A bit sad that we have to do that manually..

garrett:
Yeah, all the new underling stuff should just work in WebKit too. But sometimes a browser drifts behind another. (It's WebKit lately, but they've gotten much better over the past year and a half. Chrome lags sometimes too. And sometimes Firefox and WebKit intentionally drag their feet behind Chrome, due to privacy concerns, like with WebUSB.) :shrug: 

At least I checked it in WebKit, found the problem, and had a solution. It's sad they don't support the shorthand yet, but at least splitting it out to all the different tweaks works.

(WebKit also doesn't let one avoid skipping ink, but Apple has been known for underlines skipping ink for _decades_. Now the other browsers and platforms have caught up and do it too... but at least Firefox and Chrome allow you to turn it off via CSS. It's nice by default, but having URL text as a link is a little messy looking.)

PR #398, opened 2021-04-13
Frontpage screenshot navigation

garrett:
Screenshot navigation now includes mouse and keyboard with this PR.

Swipes are currently not included, as they're more complex and browsers do different things. I have a general idea how to implement it, and it wouldn't be too hard, but I want to get this in first.

The mouse support has huge target areas so it's easy to tap to the next. This means mobile should be decent enough already, even if swipes aren't supported yet. It's also good for everyone using a mouse, as they don't have to click on the arrow. The arrow does light up when hovering and there is a pointing cursor. Clicking elsewhere will dismiss the popup (as before).

Keyboard support includes arrow keys, vi navigation, and also WASD because: why not?


garrett:
Oh, I realized I did not implement page up/down... whoops. Probably fine enough without, but it'd be easy to add.

garrett:
Preview @ https://garrett.github.io/cockpit-project.github.io/

allisonkarlitskaya (commented):
I feel like maybe the arrows should disappear when the mouse isn't moving?

garrett:
> I feel like maybe the arrows should disappear when the mouse isn't moving?

There's also the keyboard navigatability of it as well. It shows that there are more pictures before and after.

Also: I've added page up/down, space, shift+space, and backspace for sake of completion. This should cover just about every way someone might expect to navigate using a keyboard (except home/end, as I only have next/previous support).

garrett:
Hmm, I could play around with fading out the arrows completely when hovering the picture. And perhaps override that when the keyboard has been used? I'm just worried about discoverability.

garrett:
BTW: Here's a visualization of the active click areas for next/previous:

Screenshot_2021-04-13 Cockpit Project — Cockpit Project

If you click outside that zone, then it'll dismiss the zoomed view. This works much like a touchscreen kindle.

(The top and bottom are intentionally escape zones, as is the main part of the picture. But these HUGE link areas for next/previous ensure someone doesn't misclick/mistouch and dismiss the zoomed view.)

assets/site.sass
```diff
@@ -404,6 +410,42 @@ p.screenshot img
     text-shadow: 1px 1px 3px #000
     animation: 300ms ease-in-out fadeIn
     text-align: center
+    @media (prefers-reduced-motion: reduce)
+      animation: none
+
+  a
```
KKoukiou, line 416: Are you measing a { here?

garrett: No, this is SASS, not SCSS

garrett: I've used SASS on the website for years; was considering porting to SCSS, but it's basically the same except mandatory indentations and lack of {;} chars.

garrett: This covers the differences: https://sass-lang.com/documentation/syntax

garrett:
OK, I hammered out touch support for swiping. It works with touch emulation in the browser, but I haven't tested it on a real device yet.

garrett:
Touch support confirmed to work on a phone, added page scroll locking, and as the mouse wheel doesn't work for scrolling, I've rebound it to scroll through the gallery, as it was super-simple to add (similar to swiping).

I've only added scroll support for up/down, but not left/right. Most mice don't even support left/right anyway. The direction is bound to the direction of scrolling the page. That is, the direction you use to scroll the page down will advance to the next. Natural scrolling vs. ...uh... "unnatural" (???) scrolling will flip the direction of page scrolling and therefore switch the direction of gallery cycling too. (In other words, it should work as expected for everyone.)

I've merged it all back into the original commit, so it's squashed down to two commits. The screenshot navigation and a reduced motion one (which is mostly unrelated, but I added some reduced motion for screenshot nav and figured it would make sense elsewhere too).

garrett:
I split out the JS into its own file, finally. It grew too much to really have JS in HTML in Markdown. :wink: 

Also: I'm not addressing the overlapping arrows here in this PR. That should be a follow-up.

allisonkarlitskaya (changes requested):
I don't like the swipe gesture on mobile.  It's too jarring.

If we can't implement a better transition (like sliding through the images) then I'd prefer if we didn't have it at all.  Isn't there some component somewhere that we could use for this?

garrett:
Removed swiping for now and force-pushed.

At least, for mobile, the target areas to tap are quite large and flush to the edges. (This helps desktop too, but is critical for mobile, especially as one cannot hover on mobile to have the hover affordance.)

I have an idea on how to reimplement this with CSS scroll stops (which is now well supported in all browsers, both desktop and mobile), but it would require rewriting more than I'm comfortable with doing in this PR and would further delay the ability to switch screenshots.

So I would suggest we merge this in, without swipe support for now, and I reimplment it.

It would be cool, as I haven't worked with scroll snapping yet, and that could actually be useful in Cockpit in places as well.

allisonkarlitskaya (approved):
Thanks!

PR #404, opened 2021-04-22
apps: Add application data & page

### LINKED ISSUE #348: highlight add-ons for Cockpit
I don't think the website talks about https://github.com/Scribery/cockpit-session-recording, for example. We probably want to highlight Cockpit Podman a bit more as well.

Additionally, if we had instructions on how to add these in the various distros (such as adding `cockpit-session-recording` on Fedora), that would be fantastic.

garrett:
Includes links to the page as well.

Fixes #348.

### DIFF
--- a/_data/applications.yml
+++ b/_data/applications.yml
@@ -1,0 +1,175 @@
+# Entries on the page show up in the order listed here, but with
+# official: true (official) first
+# official: nil (unofficial; non-team RH developed) second (this is omission of `official`),
+# official: false (third party) last
+
+# The Cockpit group is special, it inherits source, issues, license for the `source: cockpit` entries
+cockpit:
+  source: https://github.com/cockpit-project/cockpit/
+  issues: https://github.com/cockpit-project/cockpit/issues
+  license: LGPL 2.1
+  comment: |
+    This is the default template that "source: cockpit" pulls from, so source, issues, and license don't need to be filled out everywhere.
+
+storaged:
+  title: Storage
+  package: cockpit-storaged
+  source: cockpit
+  official: true
+  description: |
+    Manage your system's storage. Supports local partitions, encryption, NFS, RAID, iSCSI, and more.
+
+networkmanager:
+  title: Network
+  package: cockpit-networkmanager
+  source: cockpit
+  official: true
+  description: |
+    Manage your network interfaces and edit your firewall with ease.
+
+packagekit:
+  title: Software Updates (PackageKit)
+  package: cockpit-packagekit
+  source: cockpit
+  official: true
+  description: |
+    See and apply updates to your system. Supports RPM and DEB based systems through PackageKit.
+
+ostree:
+  title: Software Updates (OSTree)
+  package: cockpit-ostree
+  official: true
+  source: https://github.com/cockpit-project/cockpit-ostree
+  issues: https://github.com/cockpit-project/cockpit-ostree/issues
+  license: LGPL 2.1
+  description: |
+    Upgrade OSTree-based systems in your browser.
+
+machines:
+  title: Virtual Machines
+  package: cockpit-machines
+  source: https://github.com/cockpit-project/cockpit-machines
+  issues: https://github.com/cockpit-project/cockpit-machines/issues
+  license: LGPL 2.1
+  official: true
+  description: |
+    Create, run, and manage virtual machines in your browser.
+
+podman:
+  title: Podman Containers
+  package: cockpit-podman
+  source: https://github.com/cockpit-project/cockpit-podman
+  issues: https://github.com/cockpit-project/cockpit-podman/issues
+  license: LGPL 2.1
+  official: true
+  description: |
+    Download, use, and manage containers in your browser. (Podman replaces Docker.)
+
+selinux:
+  title: SELinux
+  package: cockpit-selinux
+  source: cockpit
+  official: true
+  description: |
+    View and manage SELinux exceptions. Instantly apply overrides and export your rules to apply to other systems.
+
+kdump:
+  title: Kernel Dump
+  package: cockpit-kdump
+  source: cockpit
+  official: true
+  description: |
+    Debugging the Linux kernel? This `kdump` add-on helps with catching stack traces.
+
+composer:
+  title: Cockpit Composer
+  package: cockpit-composer
+  source: https://github.com/osbuild/cockpit-composer/
+  issues: https://github.com/osbuild/cockpit-composer/issues
+  description: |
+     Composer is an interface for OSBuild that generates custom images suitable for deploying systems or uploading to the cloud. Images built with Composer are for RPM-based distributions such as Fedora, Red Hat Enterprise Linux, and CentOS.
+
+certificates:
+  title: Certificate Management
+  package: cockpit-certificates
+  source: https://github.com/skobyda/cockpit-certificates
+  issues: https://github.com/skobyda/cockpit-certificates/issues
+  prerelease: true
+  description: |
+    Certificate management app for Cockpit.
+
+sosreport:
+  title: Diagnostic Reports
+  package: cockpit-sosreport
+  source: cockpit
+  official: true
+  description: |
+    Collect system configuration and diagnostic information from systems, to help with diagnosing problems.
+
+389-ds:
+  title: 389 Directory Server
+  package: cockpit-389-ds
+  homepage: https://directory.fedoraproject.org/docs/389ds/download.html
+  source: 
+  description: |
+    A web-based interface to the enterprise-class Open Source LDAP server for Linux.
+
+session-recording:
+  title: Session Recording
+  package: cockpit-session-recording
+  source: https://github.com/Scribery/cockpit-session-recording
+  issues: https://github.com/Scribery/cockpit-session-recording/issues
+  description: |
+    Session recording configuration and playback for Cockpit, based on [Scribery](https://scribery.github.io/), to record a user's actions for watching later.
+
+subscriptions:
+  title: Subscription Manager
+  package: cockpit-subscriptions
+  source: https://github.com/candlepin/subscription-manager
+  issues: 
+  description: |
+    Manage subscriptions to Red Hat Enterprise Linux from your web browser.
+
+# leapp:
+#   package: cockpit-leap
+#   title: Leapp
+#   homepage: https://leapp.readthedocs.io/en/latest/
+#   source: https://github.com/oamg/leapp
+#   license: Apache 2
+#   issues: https://github.com/oamg/leapp/issues
+#   description: |
+#     Leapp is an OS and application modernization framework. It is used to upgrade Red Hat Enterprise Linux. (For example: Upgrading RHEL 7 to 8.)
+
+ovirt-dashboard:
+  title: oVirt Dashboard
+  package: cockpit-ovirt-dashboard
+  source: https://github.com/oVirt/cockpit-ovirt
+  issues: https://bugzilla.redhat.com/buglist.cgi?product=cockpit-ovirt&resolution=---
+  description: |
+    oVirt manages Virtual Machines (VMs) in a datacenter/cluster. Scales easily from tens to tens of thousands VMs running on multiple KVM hypervisor hosts.
+
+zfs:
+  title: ZFS Manager
+  package: cockpit-zfs
+  source: https://github.com/optimans/cockpit-zfs-manager
+  issues: https://github.com/optimans/cockpit-zfs-manager/issues
+  official: false
+  prerelease: true
+  description: |
+    An interactive ZFS on Linux admin package for Cockpit.
+
+samba:
+  title: Samba Manager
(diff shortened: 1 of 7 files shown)


garrett:
Preview should be at https://garrett.github.io/cockpit-project.github.io/applications.html when GitHub builds it.

garrett:
BTW: The link on the applications page to edit it won't work until this PR is merged.

garrett:
Anyway, this is meant to be a good first step. We can improve it in additional PRs too.

garrett:
Oh, I should change the link to the YAML file.

garrett:
Fixed! The edit link on the page now points to the applications.yml file in the _data/ subdirectory. It still won't work until after this PR is committed. :wink:

garrett:
Note: This PR is hilariously 404 by sheer coincidence. :rofl:

_data/applications.yml
```diff
@@ -0,0 +1,151 @@
...
+  title: Machines
+  source: https://github.com/cockpit-project/cockpit-machines
+  issues: https://github.com/cockpit-project/cockpit-machines/issues
+  package: cockpit-machines
+  official: yes
+  license: LGPL 2.1
+  description: |
+    Create, run, and manage virtual machines in Cockpit.
+
+podman:
```
marusak, line 18: This one misses `package: cockpit-podman`

_data/applications.yml
```diff
@@ -0,0 +1,151 @@
...
+  official: yes
+  package: cockpit-sosreport
+  source: cockpit
+  description: |
+    A tool to collect system configuration and diagnostic information from systems, to help with diagnosing problems.
+
+389-ds:
+  title: 389 Directory Server
+  package: cockpit-389-ds
+  description: |
```
marusak, line 106: This one lives here: https://github.com/389ds/389-ds-base

marusak:
Nice one!

Here a few of my comments:
1. I don't think we need to show license. 99% of users don't really care. Those who do can surely find it easily, especially since we point them to repositories. Now it just takes space in my option.
2. Issue tracker is tricky. There is upstream and bunch of downstreams. Like for cockpit-podman, you can report it to github, fedora bugzilla, rhel bugzilla, debian issue tracker... And I would say that unless you are running the newest version you should be reporting downstream. So pointing to upstream issue tracker might have different effect that we would like to see. I think, that if we just add 'Project home' or something and point it to github, users will be able to open issue in github if they want to and if they are used to report to downstream they will proceed that way.
3. I like how it looks! I like badges, descirption and all :) But at the same time I find it very hard to read and just get the idea of available apps.  I would expect to just take a look at the page and immediately see app names. Now I need to scan through the page, parse all the content just to find names. Table or list would be much easier to navigate. Or maybe make titles to stand out more?

garrett:
Sure! This is just a first pass. We can fix it up later.

I can address these issues on Monday. Thanks for the feedback!

marusak:
There are two samba clients:
https://github.com/enira/cockpit-smb-plugin
https://github.com/45Drives/cockpit-samba-manager

I would say that the latter one is active and more advanced. The first one is not dead either though.

_data/applications.yml
```diff
@@ -0,0 +1,159 @@
...
+    Upgrade OSTree-based systems in your browser.
+
+composer:
+  title: Cockpit Composer
+  package: cockpit-composer
+  source: https://github.com/osbuild/cockpit-composer/
+  issues: https://github.com/osbuild/cockpit-composer/issues
+  description: |
+     Composer generates custom images suitable for deploying systems or uploading to the cloud. It integrates into Cockpit as a frontend for osbuild composer.
```
KKoukiou, line 90: It created only rpm based images, shall we be more specific here?

_data/applications.yml
```diff
@@ -0,0 +1,159 @@
...
+  description: |
+    A web-based interface to the enterprise-class Open Source LDAP server for Linux.
+
+session-recording:
+  package: cockpit-session-recording
+  title: Session Recording
+  source: https://github.com/Scribery/cockpit-session-recording
+  issues: https://github.com/Scribery/cockpit-session-recording/issues
+  description: |
+    Ssession recording configuration and playback for Cockpit.
```
KKoukiou, line 123: ssession

KKoukiou (changes requested):

applications.md
```diff
@@ -0,0 +1,24 @@
+---
+title: Applications
+---
+
+The Cockpit Web Console is extendable. The Cockpit team and many other projects have built applications that are easy to install.
```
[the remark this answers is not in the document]
garrett: ...?

What's Composer then? It's an app built by another project within Red Hat.

We could s/project/team/; same difference in this context.

_data/applications.yml
```diff
@@ -0,0 +1,159 @@
...
+  description: |
+    Download, use, and manage containers in your browser. (Podman replaces Docker.)
+
+selinux:
+  title: SELinux
+  official: true
+  source: cockpit
+  package: cockpit-selinux
+  description: |
+    View and manage SELinux exceptions inside Cockpit. Instant apply overrides and export your rules to apply to other systems.
```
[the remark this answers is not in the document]
garrett: Instant-apply.

But it could be "instantly apply" instead.

_data/applications.yml
```diff
@@ -0,0 +1,159 @@
...
+  description: |
+    Manage your system's storage. Supports local partitions, encryption, NFS, RAID, iSCSI, and more.
+
+kdump:
+  title: Kernel Dump
+  official: true
+  source: cockpit
+  package: cockpit-kdump
+  description: |
+    Debugging the Linux kernel? Cockpit's kdump add-on helps catch stack traces.
```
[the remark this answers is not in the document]
garrett: It's "helps to catch"; "to" is implied. It could also be "helps with catching".

_data/applications.yml
garrett: Uh...? It creates images for VMs, not RPMs, right? I thought I copied this one though.

_data/applications.yml
garrett: :grin:

_data/applications.yml
garrett: Yep, copy/pasted from https://github.com/osbuild/cockpit-composer/#cockpit-composer

garrett:
I've pushed up several changes, as suggested.

I still need to fix the mobile support, as moving it to tables totally broke smaller widths. But, I'll do that tomorrow. It's getting late here already.

Aside from the mobile support, what do you think?

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+  package: cockpit-machines
+  official: true
+  license: LGPL 2.1
+  description: |
+    Create, run, and manage virtual machines in Cockpit.
+
+podman:
+  title: Podman Containers
+  source: https://github.com/cockpit-project/cockpit-podman
+  issues: https://github.com/cockpit-project/cockpit-podman/issues
```
martinpitt, line 27: package: cockpit-podman

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+
+machines:
+  title: Virtual Machines
+  source: https://github.com/cockpit-project/cockpit-machines
+  issues: https://github.com/cockpit-project/cockpit-machines/issues
+  package: cockpit-machines
+  official: true
+  license: LGPL 2.1
+  description: |
+    Create, run, and manage virtual machines in Cockpit.
```
martinpitt, line 22: I'd leave out the "in Cockpit" -- this is all about Cockpit, after all?

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+
+podman:
+  title: Podman Containers
+  source: https://github.com/cockpit-project/cockpit-podman
+  issues: https://github.com/cockpit-project/cockpit-podman/issues
+  official: true
+  license: LGPL 2.1
+  description: |
+    Download, use, and manage containers in your browser. (Podman replaces Docker.)
```
martinpitt, line 31: Likewise, the "in your browser" is redundant. Or it should at least be "in Cockpit" for consistency, depending on the resolution of the previous item.

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+  description: |
+    Certificate management app for Cockpit.
+
+sosreport:
+  title: Diagnostic Reports
+  official: true
+  package: cockpit-sosreport
+  source: cockpit
+  description: |
+    A tool to collect system configuration and diagnostic information from systems, to help with diagnosing problems.
```
martinpitt, line 107: I'd drop "A tool to"

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+  source: https://github.com/candlepin/subscription-manager
+  issues: 
+  description: |
+    Manage subscriptions to Red Hat Enterprise Linux from your web browser.
+
+leapp:
+  package: cockpit-leap
+  title: Leapp
+  homepage: https://leapp.readthedocs.io/en/latest/
+  source: https://github.com/oamg/leapp
```
martinpitt, line 137: Somehow this does not appear on the rendered page.

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+  license: Apache 2
+  issues: https://github.com/oamg/leapp/issues
```
martinpitt, line 139: Neither does this "issues:" link

_data/applications.yml
```diff
@@ -0,0 +1,175 @@
...
+subscriptions:
+  package: cockpit-subscriptions
+  title: Subscription Manager
+  source: https://github.com/candlepin/subscription-manager
+  issues: 
+  description: |
+    Manage subscriptions to Red Hat Enterprise Linux from your web browser.
+
+leapp:
+  package: cockpit-leap
```
martinpitt, line 134: "leap" is certainly a typo, but even "cockpit-leapp" does not exist in neither Fedora nor RHEL 8.5. Nor does "leapp" or anything relevant containing "leap". The [spec](https://github.com/oamg/leapp/blob/master/packaging/leapp.spec) does not have any cockpit bits. So IMHO it's too unclear in the current state how to even use/find this, and we should comment it out for now.

martinpitt (changes requested):
Thanks! I really like the declarative style here.

_data/applications.yml
garrett: Yeah, both should be in your browser, probably. They're things that one usually doesn't do in a browser, so it's notable. I do agree that as they're on the Cockpit website and are called cockpit-foo as packages, it's redundant to say Cockpit. But browser is still notable.

_data/applications.yml
garrett: Good call.

_data/applications.yml
garrett: It has to have a title and description, I think. Perhaps I forgot the description, although I thought I wrote it. I'll re-check.

garrett: Oh you mean source. That used to be there. It was asked to be removed. But that's before I changed the layout.

_data/applications.yml
garrett: They did. They were asked to be removed earlier.

_data/applications.yml
garrett: Then we need to remove it. And can comment it out for now, as you suggest.

garrett:
@martinpitt: Thanks for your feedback! I'll implement the changes now.

garrett:
Suggestions applied.

I'll work on fixing mobile now. (It'll probably look the same.)

garrett:
Squashed down to 1 commit.

martinpitt (approved):
Super cool, many thanks for putting this together!

marusak:
Very nice! Like it much more than the first version, thanks for working on it :)

I still think that if `Issue tracker` is just `<source>/issue` it is not necessary. Just when they have disabled issues on github, then it is good to point to other issue tracker (like for [ovirt](https://github.com/oVirt/cockpit-ovirt)).

_data/applications.yml
marusak: This is still not added. Can you add it as `Source`?

garrett:
> I still think that if Issue tracker is just <source>/issue it is not necessary. Just when they have disabled issues on github, then it is good to point to other issue tracker (like for ovirt).

I removed it at first, but added it back, because I think there's room for it now _and_ is useful to have. (I've found links to issues on flathub, which is also a gallery of apps that can be installed, similarly useful to report bugs.)

Perhaps I could/should change the URL text into actual words though and drop the headings. Doing so would make it more compact and less distracting. But we can change that later too, in a follow-up.

garrett:
Thinking about a tweet:

---

You can do a lot more with Cockpit by installing applications, including:

Storage (local, NFS, RAID, iSCSI), network (and firewall), updates, VMs, containers, building OS images, LDAP, and a lot more.

See the details on our new Applications page:
https://cockpit-project.org/applications.html

---

WDYT?

PR #405, opened 2021-04-28
Cockpit 243

martinpitt:
Preview: https://martinpitt.github.io/cockpit-project.github.io/blog/cockpit-243.html


_posts/2021-04-28-cockpit-243.md
```diff
@@ -0,0 +1,38 @@
...
+
+Service units show their current memory usage and which socket unit activated them (if any). The static relationships like "Requires:" are now collapsed by default, to take less space and make it easier to see the journal output below.
+
+![Service memory usage](/images/services-mem-usage.png)
+
+## Developer API: Watch for file changes without reading
+
+You can use the [cockpit.file.watch() API](https://cockpit-project.org/guide/latest/cockpit-file.html#cockpit-file-notify) to update your Cockpit page when a file changes. So far this has always automatically read the file as well, and the called handler function returned the contents.
+
+This is impractical for very large files (such as databases) or other cases where the contents is unimportant. You can now set an option `read: false` to disable the automatic reading; then the callback's `content` will always be `null`.
```
garrett, line 29:
```suggestion
Reading files in impractical in many cases, as contents may be unimportant or a file could be large (which happens with databases). Therefore, the watch API has gained an option, `read: false`, to disable the automatic reading. When read is set to false, the callback's `content` will always be `null`.
```

garrett (changes requested):
Lots of suggestions. :wink:

_posts/2021-04-28-cockpit-243.md
```diff
@@ -0,0 +1,38 @@
...
+
+Here are the release notes from Cockpit version 243.
+
+## Services: Show sockets and memory usage
+
+Socket units show the paths or ports on which they listen, and which service unit they activate.
+
+![Socket unit listening path](/images/services-socket-listen.png)
+
+Service units show their current memory usage and which socket unit activated them (if any). The static relationships like "Requires:" are now collapsed by default, to take less space and make it easier to see the journal output below.
```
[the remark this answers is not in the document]
martinpitt: "which socket unit which activated them" sounds weird, the second "which" should just go. I'll do that when squashing.

_posts/2021-04-28-cockpit-243.md
```diff
@@ -0,0 +1,38 @@
...
+
+![Service memory usage](/images/services-mem-usage.png)
+
+## Developer API: Watch for file changes without reading
+
+You can use the [cockpit.file.watch() API](https://cockpit-project.org/guide/latest/cockpit-file.html#cockpit-file-notify) to update your Cockpit page when a file changes. So far this has always automatically read the file as well, and the called handler function returned the contents.
```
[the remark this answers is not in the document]
martinpitt: s/updates/can be used to update/, it doesn't do anything by itself. I'll do that after commiting your suggestion.

_posts/2021-04-28-cockpit-243.md
martinpitt: s/in/is/, and /which happens with/for example a/

martinpitt:
Thanks @garrett! I applied your suggestions and the bunch of fixups on top. PTAL?

_posts/2021-04-28-cockpit-243.md
```diff
@@ -0,0 +1,40 @@
...
+summary: Cockpit with Services page improvements
+---
+
+Cockpit is the [modern Linux admin interface](https://cockpit-project.org/).  We release regularly.
+
+Here are the release notes from Cockpit version 243.
+
+## Services: Show sockets and memory usage
+
+Socket units show listening paths and ports as well as service units they activate.
```
garrett, line 17: So an extra word is not needed before service, but having any or the makes it clear service units is a noun, not "service (verb) units".

```suggestion
Socket units show active paths and ports, as well as any service units they activate.
```
```suggestion
Socket units show active paths and ports, as well as the service units they activate.
```

I think any works better here, but the also works. (I might be wrong, and I don't care which one you pick.)

It's also a bit difficult to read as it feels like it bounces around and keeps going on, so I added a comma. I might be wrong about this too. But "x and y as well as z" feels like a roller coaster.

_posts/2021-04-28-cockpit-243.md
```diff
@@ -0,0 +1,40 @@
...
+
+Less important static relationships like "Requires:" are collapsed by default, making it easier to see the journal output.
+
+![Service memory usage](/images/services-mem-usage.png)
+
+## Developer API: Watch for file changes without reading
+
+The [cockpit.file.watch() API](https://cockpit-project.org/guide/latest/cockpit-file.html#cockpit-file-notify) can be used to update a Cockpit page when files change. This has always read the file and returned the contents in a handler function.
+
+Reading files is impractical in many cases, as contents may be unimportant or a file could be large (for example a database). Therefore, the watch API has gained a `read` option to disable the automatic reading. When `read` is set to false, the callback's `content` will always be `null`.
```
garrett, line 31:
```suggestion
Reading files is impractical in many cases, as contents may be unimportant or a file could be large (for example a database). Therefore, the watch API has gained a `read` option to disable the automatic reading. When `read` is set to `false`, the callback's `content` will always be `null`.
```

Pretty sure false is a value too... right?

garrett (changes requested):
Minor things... looking pretty good overall.

martinpitt:
All suggestions applied, and added bodhi URLs. Ready for landing.

PR #408, opened 2021-05-06
Pixel testing blog

mvollmer:
Preview: https://mvollmer.github.io/cockpit-project.github.io/blog/pixel-testing.html

### DIFF
--- a/_posts/2021-05-06-pixel-tests.md
+++ b/_posts/2021-05-06-pixel-tests.md
@@ -1,0 +1,199 @@
+---
+title: Testing all the pixels
+author: mvo
+date: 2021-05-06 10:00
+tags: cockpit
+slug: pixel-testing
+category: tutorial
+comments: 'true'
+---
+
+The Cockpit integration tests can now contain "pixel tests". Such a
+ test will take a screenshot with the browser and compare it with a
+ reference.  The idea is that we can catch visual regressions much
+ easier this way than if we would hunt for them in a purely manual
+ fashion.
+
+## Preparing a repository for pixel tests
+
+A pixel test will take a screenshot of part of the Cockpit UI and
+compare it with a reference.  Thus, these reference images are
+important and play the biggest role.
+
+A large part of dealing with pixel tests will consequently consist of
+maintaining the reference images.  At the same time, we don't want to
+clog up our main source repository with them.  While the number and
+size of the reference images at any one point in time should not pose
+a problem, we will over time accumulate a history of them that we are
+afraid would dominate the source repository.
+
+Thus, the reference images are not stored in the source repository.
+Instead, we store them in an external repository that is linked into
+the source repository as a
+[submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules).
+That external repository doesn't keep any history and can be
+aggressively pruned.
+
+Developers are mostly isolated from this via the new
+`test/common/pixel-tests` tool.  But if you are familiar with git
+submodules, there should be no suprises for you here.
+
+A source repository needs to be prepared before it can store reference
+images in a external storage repository. You can let the tool do it by
+running
+
+```
+$ ./test/common/pixel-tests init
+```
+
+and then committing the changes it has done to the source repository.
+(Those changes will be a new or modified .gitmodules file, and a new
+gitlink at test/reference.)
+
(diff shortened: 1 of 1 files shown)


_posts/2021-05-06-pixel-tests.md
```diff
@@ -0,0 +1,191 @@
...
+maintaining the reference images.  At the same time, we don't want to
+clog up our main source repository with them.  While the number and
+size of the reference images at any one point in time should not pose
+a problem, we will over time accumulate a history of them that we are
+afraid would dominate the source repository.
+
+Thus, the reference images are not stored in the source repository.
+Instead, we store them in an external repository that is linked into
+the source repository as a submodule.  That external repository
+doesn't keep any history and can be aggressively pruned.
```
martinpitt, line 33: Possibly link to https://git-scm.com/docs/git-submodule and/or https://git-scm.com/book/en/v2/Git-Tools-Submodules ?

We don't necessarily need to explain it here, but what happens if we prune the external repo and check out an old version of cockpit.git? I.e. the referenced SHA does not exist? Or does the mere existence of a SHA reference in cockpit.git count keep all the SHAs in the github repo alive? In the latter case, the focus would shift from "don't accumulate cruft on github.com" to "we don't want to check out the entire image history when cloning cockpit.git".

_posts/2021-05-06-pixel-tests.md
```diff
@@ -0,0 +1,191 @@
...
+and then committing the changes it has done to the source repository.
+(Those changes will be a new or modified .gitmodules file, and a new
+gitlink at test/reference.)
+
+## Adding a pixel test
+
+To add a pixel test to a test program, call the new "assert_pixels"
+function of the Browser class:
+
+     ...
```
martinpitt, line 56: I suggest to drop the ... (they don't really provide much), or at least add a `def testSomeDialog(self):` header as a context.

You can use a ```py wrapper instead of indentation to get this printed more nicely.

_posts/2021-05-06-pixel-tests.md
```diff
@@ -0,0 +1,191 @@
...
+
+As mentioned above, the `test/reference/` directory is very special
+indeed, and needs to be carefully managed with the
+`test/common/pixel-tests` tool (or even more carefully with `git
+submodule` et al).
+
+First, make sure `test/reference` is up-to-date:
+
+```
+$ ./test/common/pixel-tests pull
```
martinpitt, line 76: To clarify: You need to do that after a `git pull`/`git rebase` of the main repo? I.e. this updates test/reference/ after the main repo's gitlink changes?

martinpitt, line 76: Drop the `$` for consistency with the other commands above.

martinpitt (changes requested):
Thanks! It's nice to have this documentation. I'd like to see it also in cockpit.git test/common/README.md, or README.pixel-tests.md so that it can co-evolve with the code -- but it's certainly fine to start as a blog post.

Some little suggestions.

_posts/2021-05-06-pixel-tests.md
mvollmer: > We don't necessarily need to explain it here, but what happens if we prune the external repo and check out an old version of cockpit.git?

The build will likely fail where it calls `pixel-tests pull`.  If we get past that, each assert_pixels call will fail because the expected reference image isn't there.

This is not great... should we add a explicit knob to turn off pixel tests? Like TEST_PIXELS=0 in the envrionment?

mvollmer: > Possibly link 

Yes!

_posts/2021-05-06-pixel-tests.md
mvollmer: I have added the `def`.

_posts/2021-05-06-pixel-tests.md
mvollmer: Yes, exactly.

mvollmer: Hmm, but they all have a `$`.  Except the first, which I have now fixed.

martinpitt (approved):
Danke!

PR #412, opened 2021-05-19
running: Add Safari iOS workaround

garrett:
Preview:
- text under the Safari icon: https://garrett.github.io/cockpit-project.github.io/running.html
- instructions for iPhone users: https://garrett.github.io/cockpit-project.github.io/running/safari.html

(You may need to shift-reload the page to get the new CSS when previewing.)


running/safari.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Use Cockpit with Safari on an iPhone or iPad
+---
+
+Safari on iOS and iPadOS 14.5 has experimental support for secure WebSocket sessions, which Cockpit needs to function.
+
+Unfortunately, proper WebSocket support is currently turned off by in 14.5. Until Apple makes this the default, here are instructions for a workaround to make iPhones work with Cockpit.
```
jonaharagon, line 7: My understanding that this is for self-signed/invalid certificates only? If so you should probably lay out *all* the solutions, since something like Let's Encrypt is still probably a better idea. Something like:

```suggestion
Safari's default WebSocket framework on iOS, iPadOS, and ARM-based Macs does not support the WebSocket Secure sessions which Cockpit needs to function when using a self-signed or otherwise invalid certificate, even when they are manually marked as trusted. [Installing a valid certificate]({{ site.baseurl }}/guide/latest/guide.html) should resolve this issue.

As an alternative workaround, you can enable Apple's future/experimental WebSocket framework which does work as expected:
```

garrett: Oh, right! Thanks for the catch!

I was so eager to publish the workaround, I forgot to mention that this is for self-signed certificates.

Thanks for the wording; I've applied the suggestion.

running/safari.md
```diff
@@ -2,9 +2,9 @@
 title: Use Cockpit with Safari on an iPhone or iPad
 ---
 
-Safari on iOS and iPadOS 14.5 has experimental support for secure WebSocket sessions, which Cockpit needs to function.
+Safari's default WebSocket framework on iOS, iPadOS, and ARM-based Macs does not support the WebSocket Secure sessions which Cockpit needs to function when using a self-signed or otherwise invalid certificate, even when they are manually marked as trusted. [Installing a valid certificate]({{ site.baseurl }}/guide/latest/guide.html) should resolve this issue.
```
martinpitt, line 5: I suggest to directly link to /https.html instead, i.e. https://cockpit-project.org/guide/latest/https.html -- then it's much easier to find.

Seems a bit odd that I don't see this second "Update running.." commit on your [preview](https://garrett.github.io/cockpit-project.github.io/running/safari.html)? Did you have to sync this manually, as this is not coming from your fork's main branch?

martinpitt (changes requested):
Amazing, thanks to both of you for figuring this out! Just a minor detail.

running/safari.md
garrett: Oh, this is in a branch, so I have to force-push to main. I'll squash the changes in, change the link and force-push now.

garrett:
I've added instructions for Safari on macOS ARM/M1, edited the text a bit more, changed the URL, added in-page links, added modified/optimized screenshots for macOS, and squashed everything down.

I hope it's ready for merging now? Thanks, everyone!

martinpitt (approved):
Very cool, thank you!

PR #413, opened 2021-05-19
running: clarify macOS Safari workaround

larskarlitski:
Add missing instruction to enable the Develop menu in the settings dialog.

Just saw this and noticed an small instruction missing. Super happy this works now!


martinpitt (approved):
Danke!

PR #415, opened 2021-05-20
guide: Drop obsolete versions

martinpitt:
The /guide/ index page only advertises the 15 most recent versions
anyway, and anything older is hopelessly outdated.

Delete everything older than 230, except for 195 which is the current
version in RHEL 7.

This significantly speeds up the site build time from 38 to 9 seconds.


garrett (approved):
Yay! This is much faster.

It works here. Not only will this make local development faster, our previews should be near-instant on GitHub now, instead of waiting a minute.

PR #416, opened 2021-05-20
Update gems (2021-05-20)

garrett:
It's just routine dependency maintenance. Since our site is static it doesn't matter much, but it's still a good idea to bump up the versions of deps from time to time.

I've browsed around locally and nothing has changed on the site (as expected).


garrett:
Preview is @ https://garrett.github.io/cockpit-project.github.io/ — but there's nothing to see. :wink:

martinpitt (approved):

PR #417, opened 2021-05-20
Update IRC to libera.chat

martinpitt:


index.md
```diff
@@ -184,7 +184,7 @@ Contributing
 : [Contribution overview](https://github.com/cockpit-project/cockpit/wiki/Contributing)
 : [Get the source](https://github.com/cockpit-project/cockpit)
 : [Join the mailing list](https://lists.fedorahosted.org/archives/list/cockpit-devel@lists.fedorahosted.org/)
-: [IRC #cockpit on Freenode.net](irc://irc.freenode.net:6667/cockpit)
+: [IRC #cockpit on libera.chat](https://libera.chat/) -- [kiwiirc web client](https://kiwiirc.com/nextclient/irc.libera.chat/#cockpit)
```
garrett, line 187: This is weird in the footer. (What we have previously was also weird in the footer.)

Perhaps we need an IRC page with instructions and click-to-opt-in iframe with the web chat? It could just be a page with IRC instructions to start though.

garrett: Or it could be a block in contribute and we could link to that. (We have a contact in the privacy; it could be a copy/paste/adjust from that.)

garrett (changes requested):
I'm not a fan of the footer. We should update the contribute with information and link it.

index.md
martinpitt: Good point, updated to keep the original form for now.

garrett (approved):
Approving for now, but let's do an immediate follow-up to add the IRC info on the contribute page and link to that from the footer instead.

PR #420, opened 2021-06-16
running: Document how to enable password SSH logins on CoreOS

martinpitt:
Same update as the cockpit/ws container documentation update:
https://github.com/cockpit-project/cockpit/commit/7a289865e7a

----

Preview: https://martinpitt.github.io/cockpit-project.github.io/running.html#coreos

### DIFF
--- a/running.md
+++ b/running.md
@@ -160,18 +160,26 @@ The standard Fedora CoreOS image does not contain Cockpit packages.
 
 2. Reboot
 
-3. Run the Cockpit web service with this privileged container (as root):
+Steps 1 and 2 are enough when the CoreOS machine is only connected to through another host running Cockpit.
+
+If you want to also run a web server to log in directly on the CoreOS host:
+
+3. Enable password based SSH logins, unless you only use [SSO logins](https://cockpit-project.org/guide/latest/sso.html):
+   ```
+   echo 'PasswordAuthentication yes' | sudo tee /etc/ssh/sshd_config.d/02-enable-passwords.conf
+   sudo systemctl try-restart sshd
+   ```
+
+4. Run the Cockpit web service with a privileged container (as root):
    ```
    podman container runlabel --name cockpit-ws RUN docker.io/cockpit/ws
    ```
 
-4. Make Cockpit start on boot:
+5. Make Cockpit start on boot:
    ```
    podman container runlabel INSTALL docker.io/cockpit/ws
    systemctl enable cockpit.service
    ```
-
-_Steps 3 and 4 are optional if the CoreOS machine will only be connected to from another host running Cockpit._
 
 Afterward, use a web browser to log into port `9090` on your host IP address as usual.
 


running.md
```diff
@@ -160,18 +160,24 @@ The standard Fedora CoreOS image does not contain Cockpit packages.
...
 
-4. Make Cockpit start on boot:
+5. Make Cockpit start on boot:
    ```
    podman container runlabel INSTALL docker.io/cockpit/ws
    systemctl enable cockpit.service
    ```
 
-_Steps 3 and 4 are optional if the CoreOS machine will only be connected to from another host running Cockpit._
+_Steps 3 to 5 are optional if the CoreOS machine will only be connected to from another host running Cockpit._
```
garrett, line 180: When looking at the preview this morning, I realized it's kind of odd to have "optional" _after_ the steps. (It's a problem now, but it's even more noticeable with more points.)

Solutions (at least one or both):
- Move the optional message above the optional steps.
- State the step is optional in the step.

garrett (changes requested):
The optional parts should probably look different. I went into depth inline.

Note: When setting up Cockpit on my Fedora IoT server (also based on ostree), I attempted to follow the CoreOS instructions and they didn't work. I ended up just doing an overlay with `rpm-ostree install` and enabling Cockpit. I don't know if the CoreOS instructions are supposed to work on IoT or not.

running.md
martinpitt: Note that this was just a straight copy from https://github.com/cockpit-project/cockpit/blob/master/containers/ws/README.md . I updated the text as per your suggestions now. Once we land this, we need to sync this back to cockpit's container README, cockpit-container.git, and the docker.io/quay.io container descriptions.

martinpitt:
Out of interest, what exactly failed on Fedora IOT?

garrett:
> Out of interest, what exactly failed on Fedora IOT?

I don't exactly remember, but I think I couldn't access Cockpit. Perhaps it was firewall related issues? I'd have to try to set it up like that again. 

Meanwhile, I've been happy with a standard overlay. I want to make sure Cockpit is updated, and I'm not sure if the container approach is auto-updated. (I assume so? But the docs didn't say if it would be, or how to make sure.)

running.md
```diff
@@ -160,19 +160,25 @@ The standard Fedora CoreOS image does not contain Cockpit packages.
...
+   sudo systemctl try-restart sshd
+   ```
+
+2. Run the Cockpit web service with this privileged container (as root):
    ```
    podman container runlabel --name cockpit-ws RUN docker.io/cockpit/ws
    ```
 
-4. Make Cockpit start on boot:
+3. Make Cockpit start on boot:
```
garrett, line 176: I _think_ we can use 3. 4. 5. here still.

martinpitt: That's not how markdown enumerations work, though. You can also write "0." or "1." everywhere, and it will always do the "proper" counting from 1. I'll try, though.

martinpitt: Yup, pushed, and it's still rendered as "1. 2. 3.". Doesn't hurt though to keep it as 3. 4. 5. in the MD sources.

running.md
```diff
@@ -160,19 +160,25 @@ The standard Fedora CoreOS image does not contain Cockpit packages.
 
 2. Reboot
 
-3. Run the Cockpit web service with this privileged container (as root):
+This is sufficient if the CoreOS machine will only be connected to from another host running Cockpit. If you also want to run a web server for logging in directly:
```
garrett, line 163:
```suggestion
Steps 1 and 2 are enough when the CoreOS machine is only connected to through another host running Cockpit.

If you want to also run a web server to log in directly on the CoreOS host:
```

running.md
```diff
@@ -160,19 +160,25 @@ The standard Fedora CoreOS image does not contain Cockpit packages.
...
+
+3. Enable password based SSH logins, unless you only use [SSO logins](https://cockpit-project.org/guide/latest/sso.html):
+   ```
+   echo 'PasswordAuthentication yes' | sudo tee /etc/ssh/sshd_config.d/02-enable-passwords.conf
+   sudo systemctl try-restart sshd
+   ```
+
+4. Run the Cockpit web service with this privileged container (as root):
```
garrett, line 171:
```suggestion
4. Run the Cockpit web service with a privileged container (as root):
```

garrett (changes requested):
Looks good. I've added a few suggestions, but it's nothing major.

martinpitt:
Thanks @garrett ! Applied your suggestions.

martinpitt:
I sent https://github.com/cockpit-project/cockpit/pull/16027 to update the primary version of that file accordingly.

PR #432, opened 2021-07-06
Scripts cleanup

garrett:
~This needs to be rebased once #426 lands.~


_scripts/toolbox-run
```diff
@@ -1,3 +1,8 @@
-#!/bin/sh
+#!/bin/bash
 
-toolbox run -c cockpit-website bundle exec jekyll server $@
+if [[ -n $1 ]]
```
martinpitt, line 3: To me this is a bit unfamiliar and a bashism. The canonical/POSIX way is `if [ -n "$1" ]`. But should be ok.

martinpitt (approved):
Thanks! This is still draft, but I think approvals survive undrafting.

_scripts/toolbox-run
garrett: It's definitely a bashism: https://stackoverflow.com/questions/3427872/whats-the-difference-between-and-in-bash

I could change it to what you suggest and switch back to sh though; probably fine here.

garrett:
Not quite done with this PR, but it's getting close.

PR #433, opened 2021-07-06
Remove broken link to legacy Fedora packages services

rugk:
The Fedora packages app has been deprecated for a while for now,
see https://discussion.fedoraproject.org/t/https-apps-fedoraproject-org-packages-is-down-and-has-been-for-a-while/23662/3?u=rugk
The suggested alternative is https://src.fedoraproject.org.

Unfortunately, I also found no way to link to the cockpit packages as before. There is only the usual page and also no way to show dependencies:
https://src.fedoraproject.org/rpms/cockpit

As such I just removed the link.

### DIFF
--- a/running.md
+++ b/running.md
@@ -154,7 +154,7 @@ The standard Fedora CoreOS image does not contain Cockpit packages.
    rpm-ostree install cockpit-system cockpit-ostree cockpit-podman
    ```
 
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Makes sense, thank you for spotting this!

garrett:
Oh! We could also link to https://cockpit-project.org/applications.html instead now.

rugk:
Thx, done in https://github.com/cockpit-project/cockpit-project.github.io/pull/434 :blue_book:

PR #439, opened 2021-07-19
script: Handle lack of release notes on main repo

garrett:
I moved the title string into the frontmatter, always basing it on Cockpit's own version, allowing for Cockpit itself to have no release-note tags and yet still generate proper release notes.

(Subprojects may have release-note tags still, even if Cockpit itself does not have anything noteworthy.)

Previous version would erroneously grab the first version of something with release notes, which works fine if Cockpit itself has release notes. Whoops.


martinpitt (dismissed):
Danke!

allisonkarlitskaya (approved):

PR #441, opened 2021-07-21
Replace broken links with linux.die.net links

jpall12:
## All edits where made on the [PCP Metrics page](https://cockpit-project.org/guide/latest/feature-pcp.html) 

1. Changed `pmstat` link from [this](https://pcp.io/books/PCP_UAG/html/LE38515-PARENT.html#LE91266-PARENT) to [this](https://linux.die.net/man/1/pmstat) 
2. Changed `pminfo` link form [this](https://pcp.io/books/PCP_UAG/html/LE60452-PARENT.html) to  [this](https://linux.die.net/man/1/pmstat) 

If you feel that the updated links are not the best choice feel free to edit or drop them below.

### DIFF
--- a/guide/latest/feature-pcp.html
+++ b/guide/latest/feature-pcp.html
@@ -36,8 +36,8 @@ layout: guide
     systemd unit is running or not. The "Enable PCP metrics collector" button on the
     "Performance Metrics" page will enable and start this service.</p>
 <p>To see similar metrics data from the command line, you can use tools like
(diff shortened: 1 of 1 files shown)


marusak:
This is just sync from https://github.com/cockpit-project/cockpit (see `doc/guide/feature-pcp.xml`) so on the next sync, it would get replaced back. Are you interested in sending PR there as well or should I do it?

jpall12:
Oh I didn't realize this was just a sync, I will open a PR there. Thanks for pointing this out

PR #443, opened 2021-07-22
guide: drop old versions

allisonkarlitskaya:
And move over to a world where we manually edit index.html to preserve
old versions that we consider to be important.

From now on, the release process will only update latest/.

Also drop HEADER.html and FOOTER.html files which were probably
accidentally included as artifacts from the build process.

Note: my attempts to add a jekyll header to the toplevel index.html doesn't seem to have gotten me very far.  Probably there's an easy way to fix that, though.  What would be extra nice is if we could figure out a way to style the entire directory without the annotations and/or get the style from the website into the cockpit repo so we don't need to depend on jekyll to do that for us...

I probably also need to go through and undelete a few of the older versions that are in stable releases of various distros we care about:
 - buster → 188 (but 239 from backports)
 - 2004 → 215
 - 2104 → 238
 - 8.4 → 238


garrett:
> manually edit index.html

If it's manually edited, should we make it a Markdown file instead (and use the default template for the site)?

martinpitt (approved):
Thanks! Let's land this after https://github.com/cockpit-project/cockpit/pull/16131 to make sure new versions are not coming back. Further cleanup of the manual index page can certainly happen in a follow-up. Mixing it with this mega-commit would be rather unwieldy IMHO.

garrett:
I've moved the guide index from HTML to Markdown @ 0e8eb4c.

garrett:
I've made it so the documentation highlight is correct and force-pushed my change to 55f476a. Preview is still in the same place.

PR #452, opened 2021-09-01
Use fedora35 in generate script

skobyda:


skobyda:
Ups, accidentally added 2 commits, will drop the first one in a second

martinpitt (approved):
Thanks!

PR #455, opened 2021-09-09
script: Bump toolbox-create script to Fedora 35

garrett:
Fedora 35 beta's toolbox cannot currently run Fedora 34 containers. However, Fedora 34 and Fedora 35 both can run Fedora 35 containers and this script works perfectly well in Fedora 35, so I'm bumping the version.


martinpitt (approved):
That seems fine, thanks!

PR #472, opened 2021-11-15
Update logo to 2021 edition

garrett:
Not 100% done yet — and I may need to optimize the logo a little more too. But here's a preview of what I currently have.

Live preview: https://garrett.github.io/cockpit-project.github.io/

Before:

Screen Shot 2021-11-15 at 18 12 15

After:

Screen Shot 2021-11-15 at 18 12 30

It's quite subtle.

- Changed text to RedHatDisplay. (It's quite similar.
- Tweaked kerning optically.
- Embiggened the plane.
- Thinned down the outline to be the same weight as the text.


garrett:
This is one of those things where an organization spends 100 million bucks with a design firm and it results in a logo that only designers will notice the difference. (Minus the design firm and 100 million bucks.)

PR #481, opened 2022-01-03
scripts: Handle empty body in release-notes PRs

garrett:
I wrote the script to assume there would always be a body, but the GitHub API returns a nil and Ruby seems to like to change an empty string into a nil sometimes when match and gsub are used.

### DIFF
--- a/_scripts/generate-release-notes.rb
+++ b/_scripts/generate-release-notes.rb
@@ -204,7 +204,7 @@ def format_issue(issue, repo)
   heading_prefix = repo.match('-') ? "#{drop_prefix(repo)}: " : ''
 
   # Start polishing the issue body into release note text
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Not sure what these `&` are doing, but I trust you on that :grin: Thanks!

garrett:
These `&`s are Ruby's way of saying "ignore this function if doesn't exist", rather than throwing an error. Since strings (the class the values are supposed to be) have those functions and the `nil` class does not, we can stop short and still handle things as expected without type checking. Most of these are assignments or the statements equate to true/false (after running through `match`... but `nil` is considered false anyway too).

This kind of thing is what happens (and is sometimes needed) in languages that aren't strongly typed. :shrug:

PR #483, opened 2022-01-07
Mention adjusted install command needed for Linux Mint

queengooborg:
This PR adds a few lines to indicate that on Linux Mint, the `UBUNTU_CODENAME` variable is required rather than the `VERSION_CODENAME`.  Linux Mint defines its own version codename separate from Ubuntu's, so the backport command as-is fails.  Other Ubuntu-based distributions may do the same, but Linux Mint is probably the most important to note since it's one of the top five distributions.


martinpitt (approved):
Thanks!

PR #493, opened 2022-02-16
Update generate release notes

jelly:


_scripts/generate-release-notes.rb
```diff
@@ -84,6 +84,7 @@
 * [NAME Source Tarball](https://github.com/cockpit-project/REPO/releases/tag/VERSIONS)
 * [NAME Fedora 36](https://bodhi.fedoraproject.org/updates/?releases=F35&packages=REPO)
 * [NAME Fedora 35](https://bodhi.fedoraproject.org/updates/?releases=F34&packages=REPO)
+* [Cockpit Client](https://flathub.org/apps/details/org.cockpit_project.CockpitClient)
```
martinpitt, line 87: We don't update this every time (often there's no visible changes), but indeed it's better to leave it in the template and take it out than forget about it.

martinpitt (approved):
Cheers!

PR #498, opened 2022-02-23
Cockpit 264

garrett:
Preview: https://garrett.github.io/cockpit-project.github.io/blog/cockpit-264.html


garrett:
Since the preview shows that we probably won't have much to talk about feature wise (or a huge, notable bug fix), we'll probably just do a mid-release-cycle point release with bugfixes.

PR #502, opened 2022-03-07
Update gems to github-pages 225

martinpitt:
This will bump commonmarker to version 0.23.4 to fix
https://github.com/github/cmark-gfm/security/advisories/GHSA-mc3g-88wq-6f4x

----

I verified that the website running from local jekyll toolbox looks fine.


garrett (approved):
Everything looks correct, except for a blog summary — but that has the same issue on the deployed site. (So: It's unrelated.)

PR #503, opened 2022-03-07
blog: Improve subtitles in blog summary list

garrett:
This patch handles subheadings in a better way, by converting them to H3s (as the titles are H2s) and reformats the subheadings to be the same size as the text, with reduced space.

Before:

Screenshot 2022-03-07 at 15-36-38 Blog posts — Cockpit Project

After:

Screenshot 2022-03-07 at 15-36-25 Blog posts — Cockpit Project

### DIFF
--- a/_includes/post.html
+++ b/_includes/post.html
@@ -69,20 +69,21 @@
     | strip
     %}
     {% if skip == '' %}
+      {% assign line_stripped = line | strip_html %}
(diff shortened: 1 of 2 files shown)


martinpitt (approved):
Thanks for the tweak! Indeed that looks rather weird right now.

PR #505, opened 2022-03-17
Fix bodhi url fedora version in generation script

skobyda:
a small nitpick


martinpitt (approved):
Thanks! I'll call it "adjust" on merging though, as the current state wasn't "wrong".

PR #512, opened 2022-04-11
Drop unused events functionality and static code copies

martinpitt:
This was triggered by a dependabot alert for the recent [moment.js vulnerability](https://github.com/moment/moment/security/advisories/GHSA-8hfj-j24r-96c4)

Preview: https://martinpitt.github.io/cockpit-project.github.io/

Most of our webpage is static HTML, except for search and the gallery. I went through everything, and it works as before.


garrett:
This was all vestigial. Date handling, events, and the calendar were leftovers from when I built the website from some work I did way back in the pre-Cockpit OSAS (now OSPO) days.

This PR looks like a completely reasonable excision.

(I *think* this is the only part we're really not using in the Cockpit website?)

garrett (approved):
:+1:

PR #513, opened 2022-04-11
Cockpit 267

garrett:
Preview: https://garrett.github.io/cockpit-project.github.io/blog/cockpit-267.html

PR #515, opened 2022-04-12
Revert "Drop unused static jQuery code copy"

martinpitt:
It's apparently needed for GitHub updating the site.

This reverts commit 7d87809bc3c268d618bad8b0803a23122bba384c.

See #513


jelly:
Failure shown in CI https://github.com/cockpit-project/cockpit-project.github.io/runs/5989253060?check_suite_focus=true

garrett:
> It's apparently needed for GitHub updating the site.

It doesn't have to do with GitHub; it's needed by the site. GitHub (rightfully) can't compile and deploy Jekyll sites that have errors.

Off the top of my head, we specifically use it for the search. I wrote that years ago in CoffeeScript with jQuery. (It's probably not that hard to port it to vanilla JavaScript, but as-is, it still works.)

(BTW: Whoops. I should have caught this earlier; checking Jekyll with incremental is good for minor changes, not big ones which include file addition/deletions like this.)

garrett:
I checked and we're also using jQuery for the screenshot gallery on the front page.

It wouldn't be so hard to switch to something like [Zepto](https://zeptojs.com/) or [Cash](https://github.com/fabiospampinato/cash) (preferably Cash, as it's even smaller).

Or it could just use normal JavaScript. It wouldn't be so hard to port it, especially with a reference site from jQuery to vanilla JS, like https://youmightnotneedjquery.com/

The least amount of effort for the most gain (albeit minimal) would be using Cash... or just keep going with how it is, as it's just a static website. :wink:

PR #520, opened 2022-05-02
Add cockpit-tukit to applications

JanZerebecki:


JanZerebecki:
Fyi @skazi0

_data/applications.yml
```diff
@@ -45,6 +45,16 @@ ostree:
   description: |
     Upgrade OSTree-based systems in your browser.
 
+tukit:
```
martinpitt, line 48: Can you please move this paragraph to the end? So that we keep the official/RH/third party grouping in the file?

_data/applications.yml
```diff
@@ -45,6 +45,16 @@ ostree:
...
+  title: Software Updates (transactional or tukit)
```
martinpitt, line 49: The "or" reads a bit confusing here. (e.g. cockpit-ostree are also transactional updates). How about: "tukit transactional software updates"?

martinpitt (changes requested):
Thank you! It's really nice to see more pages! :clap:

_data/applications.yml
JanZerebecki: The software is literally called that https://github.com/openSUSE/transactional-update and tukit is only an intermediary for it like packagekit. I changed the wording a bit, hope that is better.

martinpitt (approved):
Thank you!

martinpitt:
And there it is: https://cockpit-project.org/applications.html

PR #526, opened 2022-05-18
externals: Add proxy wiki pages

### LINKED ISSUE #525: Add proxy info to externals for the website?
Right now, we don't really talk about proxies that much on the website.

Here are two pages that would be useful to show directly on cockpit-project.org:
 
- https://github.com/cockpit-project/cockpit/wiki/Proxying-Cockpit-over-NGINX
- https://github.com/cockpit-project/cockpit/wiki/Proxying-Cockpit-over-Apache-with-LetsEncrypt

There are a lot of pages that might or might not be useful. Some pages are quite old and are probably not as relevant; others are fine enough on the wiki. Yet others would be welcome on the website, where they're indexed and searchable.

It'd be easy to add and would get pulled in on the next external pages update.
https://github.com/cockpit-project/cockpit-project.github.io/blob/main/_scripts/update-external-docs.rb#L9-L14

garrett:
Adding two wiki pages to our externals, so they're on the website and indexed for easy searching.

- https://github.com/cockpit-project/cockpit/wiki/Proxying-Cockpit-over-NGINX
- https://github.com/cockpit-project/cockpit/wiki/Proxying-Cockpit-over-Apache-with-LetsEncrypt

Fixes #525

### DIFF
--- a/_scripts/update-external-docs.rb
+++ b/_scripts/update-external-docs.rb
@@ -11,6 +11,8 @@ wiki_pages = %w[
   Cockpit-Coding-Guidelines
   Contributing
   Workflow
+  Proxying-Cockpit-over-NGINX
+  Proxying-Cockpit-over-Apache-with-LetsEncrypt
 ]
 
 # Pages to fetch from the Cockpit git repo


martinpitt (approved):
I must say I'm not a fan of this copying. The wiki should be a first-class thing and also invite users to collaborate. I do get the searching bit, though.

Thanks!

PR #536, opened 2022-06-07
scripts: Switch from toolbox to podman directly

### LINKED ISSUE #535: Port toolbox scripts to use podman, possibly with the same environment as GitHub pages?
We're currently using Fedora 35 via toolbox in our scripts for running the website locally.

Should we switch to using podman containers instead? (I originally used toolbox as gems were installed inside at one point and there were permission issues with docker. Since then, I moved the gems locally to a `.gem` directory local to the checkout and podman probably solves the permission issues that hit us with docker.)

When using a podman container, we might as well use the same environment that GitHub pages uses.

garrett:
This makes the local environment:

1. More portable (as it should work whever bash & podman are).

2. Similar to GitHub Pages, as it's using the same environment.

3. Safer, as toolbox exports more of the home system than Podman does.

New commands are `container-*`. Symlinks to original `toolbox-*`
commands have been added for compatability reasons.

Fixes #535


Gemfile.lock
```diff
@@ -266,11 +260,10 @@ DEPENDENCIES
...
-  pathutil!
   webrick
 
 RUBY VERSION
-   ruby 3.0.4p208
+   ruby 2.7.3p183
 
 BUNDLED WITH
-   2.2.33
+   2.2.31
```
garrett, line 269: Likewise, GitHub Pages uses this Bundler version. (Not a big deal.)

_scripts/container-create
```diff
@@ -0,0 +1,20 @@
...
+else
+  podman="env podman"
+fi
+
+echo "## Creating website container..."
+# Use GitHub's official pages container as a base
+# but use local .gem directory (which will be mounted in /src/site)
+cat <<'EOF' | $podman build -t cockpit-website --pull -
+FROM ghcr.io/github/pages-gem
+RUN bundle config path /src/site/.gem
```
garrett, line 16: This one config change makes all the difference. Thankfully, we can pass STDIN to podman, so we don't need an actual Containerfile anywhere.

_scripts/container-create
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -e
+
+in_toolbox=/run/.containerenv
+
+if [ -f $in_toolbox ]; then
+  podman="flatpak-spawn --host env podman"
+else
+  podman="env podman"
+fi
```
garrett, line 9: This should work on a base OS or inside a toolbox (which gets passed to the host).

garrett:
Additionally, toolbx exposes $HOME and a bunch of other things, whereas Podman does not (by default, unless you tell it to), so this version of the scripts also only shares the checkout of the Cockpit website directory with the container.

(Generally, running and building Jekyll websites already has access to your computer normally. Toolbx restricted that a little. But in a Podman container via the scripts, it's _much_ more restricted. There _shouldn't_ be anything of concern, but it's always better to restrict permissions of software whenever you can.)

_scripts/container-create
martinpitt: Hah, I do [something similar](https://piware.de/gitweb/?p=bin.git;a=commitdiff;h=791a81a30e65e93bd6b10ec7ec983ec8fe79a727) :+1:

PR #541, opened 2022-06-21
scripts: Extract current user from overlay mount

garrett:
Toolbx exposed the $USER. This isn't available to a standard container.

Podman overlay mounts are shown in /etc/mtab however.

User can be inferred from overlay.

### DIFF
--- a/_scripts/generate-release-notes.rb
+++ b/_scripts/generate-release-notes.rb
@@ -2,7 +2,10 @@
 # frozen_string_literal: true
 
 # Run with:
-# _scripts/container-ruby _scripts/generate-release-notes.rb
+# _scripts/container-ruby _scripts/generate-release-notes.rb --user=me
+#
+# User name is optional. If you do not provide it, the script will attempt to
+# discover it.
 
 ### Editable variables ###
 
@@ -102,6 +105,7 @@ require 'etc'
 OptionParser.new do |opt|
   opt.on('-p', '--preview', 'Preview release notes with open PRs and more info') { @options[:preview] = true }
   opt.on('-r', '--released', 'Cockpit has been released (do not increment version)') { @options[:released] = true }
+  opt.on('-u USER', '--user=USER', 'Specify user name (instead of discovering it)') { |user| @options[:user] = user }
   opt.on('-v', '--verbose', 'Show additional information on the command line') { @options[:debug] = true }
 end.parse!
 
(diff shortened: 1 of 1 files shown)


_scripts/generate-release-notes.rb
```diff
@@ -99,6 +99,15 @@
...
 @options = OpenStruct.new
 
+# Podman mounts the filesystem as an overlay from $USER/.local/share/containers.
+# Then extract the $USER from the path.
+# This only works for user session containers. Changing to an array makes it
+# default to an empty string for system-run containers (instead of an error).
+# The Cockpit Website container is really intended to be run as a user
+# container, of course.
+mounts = File.open('/etc/mtab', 'r').read
+@user = mounts.match('/([^/]*)/.local/share/containers').to_a[1]
```
garrett, line 109: Changing the match to an actual array makes it more robust, in case there isn't a match. This can happen on system containers (which don't have any `$USER` information via this method, so there's no fallback specifically for it). When there isn't a match, `@user` defaults to an empty string (instead of an error).

martinpitt: Containers are not supposed to have any connection/knowledge about the users outside of them. These names/paths leaking through the mounts tab is more or less an accident, but not an API to rely on. Also, these names are often wrong anyway -- local user names don't necessarily correspond to the names we use on github.com or on our website (e.g. my local user name is "martin", while on our website it's "pitti").

Honestly it would be best to make this explicit with `-e user=myname` when starting the container, and avoid such hacks. If you really want to keep this, please use /proc/self/mounts -- /etc/mtab was an evolutionary dead end and has been long-obsolete.

martinpitt: Sounds good!

garrett: I can't use `-e var=val` as I'd have to adjust the script and special-case things for the variable in bash, then pass that to Podman which then would pass that to Ruby.

HOWEVER...

I already have option parsing in the Ruby script, so we can just use that.

I've updated it so that you can pass in `--user=pitti` or `-u pitti` at the end of the script. It's in the usage comments at the top too.

garrett: Ah, whoops. Still using `/etc/mtab`. I'll fix that too.

garrett: OK, we now have a manual user option and also auto-discovery using `/proc/self/mounts`. It always loads `/proc/self/mounts`; I guess I could change it to an `if`.

garrett:
OK, I think this version is good now. However, I get an error as I think I'm rate limited by GitHub. :cry:

On the plus side, I've just added better API rate limit handling as a result (in a separate commit). And now, I have to wait to see if it works (it should, still) and/or hear back from you.

PR #544, opened 2022-06-29
Make URLs into links in the HACKING.md page

stuartcampbell:
A very minor change to make a couple of the URLs into links.

### DIFF
--- a/external/source/HACKING.md
+++ b/external/source/HACKING.md
@@ -177,7 +177,7 @@ It's easy to set up your local Linux machine for rapid development of Cockpit's
 JavaScript code (in pkg/). First install Cockpit on your local machine as
 described in:
 
-https://cockpit-project.org/running.html
+<https://cockpit-project.org/running.html>
 
 Next run this command from your top level Cockpit checkout directory, and make
 sure to run it as the same user that you'll use to log into Cockpit below.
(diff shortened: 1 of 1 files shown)


martinpitt:
Hello @stuartcampbell, thanks for spotting! This page is synced automatically from https://github.com/cockpit-project/cockpit/blob/main/HACKING.md , so it needs to be changed there. Curiously GitHub renders these as actual clickable links, but not the website import. So let's use the usual `[]()` syntax in the original Markdown.

Are you interested in sending a PR there, or want me to do that? Thanks!

garrett:
I'm pretty sure you can wrap URLs in less-then and greater-than signs (`<>`) to make a URL a link, and it should work in both places. These are called "automatic" links, but they do not have any other text. It's almost always better to use different text in the standard construct than expose a URL. But development, sometimes we do want to show the URL.

Info:
<https://kramdown.gettalong.org/syntax.html#links-and-images>

(I used it in this comment to show that it works in GitHub Markdown. The URL above referrs to Kramdown, which we're using in GitHub Markdown / CommonMark mode... but it's not 100% the same. This is one of the few places where it differs (auto-linking of URLs without `<>` wrapping them.)

garrett:
Oh, this is a PR and it has this change. Apologies about this falling through the cracks!

I edited the file via GitHub directly and proposed a change in a PR over on the main repo.

PR #555, opened 2022-08-05
Add warning regarding removal of cockpit.css

joshuaboud:
I'm aware this post is known to be outdated and the starter kit should be used when getting into cockpit plugin development, but speaking from experience, this post is very enticing as a starting point (especially since going the route of [`cockpit-project.org > Contribute`](https://cockpit-project.org/external/wiki/Contributing.html)` > Further Reading > Create plugins for the user interface`, no mention of it being outdated is made). Anyone stumbling on to this page as a starting point will have some frustration saved by including a note here on cockpit.css being removed.

### DIFF
--- a/_posts/2014-11-13-cockpit-plugin-tutorial.md
+++ b/_posts/2014-11-13-cockpit-plugin-tutorial.md
@@ -82,6 +82,9 @@ The pinger tool itself looks like this:
 
 Lets take a look at the pinger HTML, and see how it works.
 
+_**Warning**: Since Cockpit 273 ([relevant pull request](https://github.com/cockpit-project/cockpit/pull/17486)), `cockpit.css` can no longer be included as described below, since it was removed from Cockpit._
+{:.warning}
+
 ```html
 <!DOCTYPE html>
 <html>


_posts/2014-11-13-cockpit-plugin-tutorial.md
```diff
@@ -82,6 +82,9 @@ The pinger tool itself looks like this:
 
 Lets take a look at the pinger HTML, and see how it works.
 
+_**Warning**: Since Cockpit 273 ([relevant pull request](https://github.com/cockpit-project/cockpit/pull/17486)), `cockpit.css` can no longer be included as described below, since it was removed from Cockpit._
```
martinpitt, line 85: This already got deprecated two years ago in [cockpit 221](https://cockpit-project.org/blog/cockpit-221.html). We have only kept the file on RHEL/CentOS 8 for backwards compatibility, but the Fedora/Debian/Ubuntu and RHEL 9 packages didn't have it any more. I think we should rather just update the blog post to drop the usage of cockpit.css, so that it stays useful.

martinpitt:
I sent PR #556 to update the tutorial to the current code, and drop cockpit.css. Does that work for you?

PR #565, opened 2022-09-28
adding SLE Micro as product that supports Cockpit

rsimai:
Hope I changed all relevant files, please let me know if anything is missing.


_data/distros.yml
```diff
@@ -48,4 +48,6 @@
 
 - slemicro:
   name: SUSE Linux Enterprise Micro
+  supported: true
+  tested: true
```
martinpitt, line 52: @garrett : Did you have some specific interpretation of "supported" and "tested" in mind? In practice, these are really the same for us.

"supported" is fungible -- We don't support it upstream, as we don't have experience with SLE Micro. However, we definitively don't test it upstream. There is the [seed of an OpenSUSE image](https://github.com/cockpit-project/bots/pull/3591), but this will require a lot of work still before the tests actually run there. So for now I'd leave this as false (or absent).

martinpitt (changes requested):
Thank you! Let's be more honest about the support status, though.

FTR, we can squash these commits when we land it.

_data/distros.yml
rsimai: I was assuming _supported_ and _tested_ comes from whoever includes Cockpit or parts of it in their distro. If so, I can assure you we're clearly testing Cockpit with SLE Micro as we offer enterprise grade support to customers.

martinpitt: @rsimai : Right, the page does not really define what these terms mean. So far these have been from upstream's point of view, but I think we need to clarify this. @garrett , WDYT?

garrett: I think it was: Officially supported by the distro, tested in our testing framework?

We can always change what's on the website to make things more clear, of course.

rsimai: Whatever you consider useful. I find it important to show that we not only include Cockpit but fully support it. Maybe there's a way to add a note "testing done by vendor" or such, if SLE Micro is an exception? Else, is there anything how we could support you in adding the required testing to your framework?

rsimai: don't want to put any pressure but since it's open for a while, can we progress this request? I'm not sure if I'm supposed to make a change, and if so, what exactly? Thanks!

garrett: Apologies for this taking so long; we had a high priority fix and I've been concentrating on dark mode too.

It's still not clear what it means, but I've been talking with @martinpitt about it in IRC. I think he can clarify. And I think it means we might need to change the page too?

garrett:
OK, from a design standpoint: The text "logo" is quite wide and it would be better if we had one that isn't as wide.

image

image

I don't have to block on this — especially as this should've been merged (in some state) at least a month ago — but it really isn't optimal. If you could provide another graphic instead that isn't so wide, we could switch to that (even as a follow-up, separate PR that would land later).

martinpitt:
I talked to @garrett : He did not have anything specific in mind wrt. the difference between "supported" and "tested". For me, these are really one and the same in upstream development: We *cannot* support OpenSUSE upstream without having a VM image (see https://github.com/cockpit-project/bots/pull/3591) and running our integration tests on it (and then fixing the heck out of the failures). This is literally how we define "support".

Also, this is the upstream project page, so this shows upstream's perspective. We do not make any claims about the support statuses of packages inside downstream distros/projects. And of course every package/update in any distro should be tested by its maintainers/uploaders -- different distros have wildly different standards there, but again that's not something we define/describe upstream.

So my proposal is to drop the "supported" vs. "tested" distinction, and merge it into a single "tested upstream" column to clarify it. And then change this PR to say `tested: false`.

rsimai:
Thanks for your responses and no worries, you're doing a great job!

I unfortunately don't have anything smaller for a logo, that's the official name of the beast. Let's go with it and target a future PR, in case I can work it out with our branding people.

I like your proposal to to resolve that tested/supported dilemma, I still wonder how we can help to have our distros tested in upstream. Reading https://github.com/cockpit-project/bots/pull/3591 I understand you effectively need to be able to download a qcow2 image of that distro with Cockpit installed and configured, that's all?

martinpitt:
Building the image will need some help how to do that, yes. Especially how to install all the test dependencies, and a `mock`/`pbuilder`/etc like environment for building the package (whatever OpenSUSE uses officially). But that's the easy part -- the hard part is to run and fix the integration tests there, which is usually a few weeks of initial work (it can be done in steps), and then an ongoing maintenance effort.

martinpitt:
@rsimai: I suggest to keep notes, decisions, and discussions around the image building in  https://github.com/cockpit-project/bots/pull/3591. Quick questions and such can/should happen on IRC in `#cockpit` (libera.chat), Imobach already contactes us there some months ago.

The supported/tested cleanup landed, so this PR is now ready to be rebased. Please set "included" to true, and "tested" to false for the time being. Let's work towards going to `true` at some point! :grin:  

Thanks!

rsimai:
I just realized this was waiting for me. Not sure if I did it right, please let me know if there's anything else.

martinpitt (approved):
Thanks! This was conflicty, so I took the liberty to rebase, deconflict, and squash your branch. I also removed the trailing empty lines.

PR #577, opened 2022-11-09
running: Clarify CentOS versions

martinpitt:
Use the same "and later" as for RHEL, to clarify that Cockpit is not only in version 7.

---

see https://github.com/cockpit-project/cockpit/issues/17891#issuecomment-1307816312


marusak (approved):
Danke :)

PR #585, opened 2022-11-24
FAQ: Initial version of frequently asked questions

### LINKED ISSUE #581: FAQ (Frequently Asked Questions)
We should probably have a FAQ. On it, we probably should have a few common issues and fixes.

- The "whilst offline" one for Ubuntu:
https://github.com/cockpit-project/cockpit/issues/16963

- The rpm-ostree issue on a system that isn't using rpm-ostree.

- Something about a white page on some devices. (Although this is mainly fixed, some iOS devices will never get updates and will never work with Cockpit.)

- Setting up a certificate with LetsEncrypt.
(We can add a link to a detailed page.)

- Setting up a proxy with NGinx
(Again, link to a full page with details.)

- How to build an app page for Cockpit
(Yet again, link to the full page.)

garrett:
Fixes #581

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -1,0 +1,141 @@
+---
+title: Frequently Asked Questions (FAQ)
+---
+
+- this list will be replaced by the Table of Contents
+{:toc}
+
+## Installation
+
+### How do I install a recent version of Cockpit on Debian or Ubuntu?
+
+While Debian and Ubuntu both ship with access to Cockpit in their main repositories, the version in each might be older than the latest available. This is due to the nature of "stable" and "long term support" distributions.
+
+There's often a newer version of Cockpit available in the "backports" repositories, for both [Debian](https://backports.debian.org/) and [Ubuntu](https://wiki.ubuntu.com/UbuntuBackports). We recommend using this newer version, if you're able to do so.
+
+Please read the installation instructions for the appropriate distribution:
+- [Debian installation instructions](running.html#debian)
+- [Ubuntu installation instructions](running.html#ubuntu)
+
+### How do I use Cockpit with a reverse proxy?
+
+Cockpit can be used from behind a proxy.
+
+Read one of the following documents:
+
+- [Proxying Cockpit over NGINX](external/wiki/Proxying-Cockpit-over-NGINX.html)
+- [Proxying Cockpit over Apache with LetsEncrypt](external/wiki/Proxying-Cockpit-over-Apache-with-LetsEncrypt.html)
+
+## Development
+
+### How do I build an app for Cockpit?
+
+- [Starter Kit](blog/cockpit-starter-kit.html)
+- [Make your Cockpit page easily installable](blog/making-a-cockpit-application.html)
+- [Additional developer-related resources are on the Contributing page](external/wiki/Contributing.html)
+
+## Troubleshooting
+
+### Login
+
+#### Blank white page after login
+
+If you see a blank page after logging in, it's a good first step to check the browser console (`Ctrl`+`Shift`+`J` in most browsers) and system logs (using `sudo journalctl --since "5 minutes ago"` on your system running Cockpit) for related error messages.
+
+##### Blank page with a Proxy
+
+You may be using Cockpit with an improperly set up proxy.
+
+###### Solutions
+{:.no_toc}
+
+- [Proxying Cockpit over NGINX](external/wiki/Proxying-Cockpit-over-NGINX.html)
+- [Proxying Cockpit over Apache with LetsEncrypt](external/wiki/Proxying-Cockpit-over-Apache-with-LetsEncrypt.html)
+
+##### Blank page using Safari
+
+Older versions of Safari on iPhones, iPads, and Safari on macOS on M1 (ARM) machines have a problem with encrypted WebSocket.
+
+###### Solution
+{:.no_toc}
+
+Newer updates of Safari no longer have this issue, but some older [devices may need to use a workaround](running/safari.html).
+
+### Software update
+
+#### Error message about "OSTree"
+
+You're seeing "OSTree is not available on this system" or "Unable to communicate with OSTree" instead of a software update page.
+
+In all likelihood, this means you have `cockpit-ostree` installed on a system that does not use `rpm-ostree` to manage its packages. You should be using Cockpit's PackageKit-based software updater in `cockpit-packagekit`, not `cockpit-ostree`.
+
+##### Solution
+{:.no_toc}
+
+Remove the ostree package:
+
+```
+dnf remove cockpit-ostree
+```
+
+`cockpit-ostree` is only intended for use on distributions that use `rpm-ostree`, such as Fedora Silverblue, Fedora Kinoite, Fedora IoT, Fedora CoreOS, and a small subset of Red Hat Enterprise Linux that uses `rpm-ostree`.
+
+#### Error message about being offline
+
+The software update page shows "packagekit cannot refresh cache whilst offline" on a Debian or Ubuntu system.
+
+##### Solution
+{:.no_toc}
+
+Create a placeholder file and network interface.
+
+1. Create `/etc/NetworkManager/conf.d/10-globally-managed-devices.conf` with the contents:
+
+    ```ini
+    [keyfile]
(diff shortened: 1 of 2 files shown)


garrett:
This needs review. I think it could stand to use some editing as well, as parts of it are probably a bit too verbose. However, it should be a good first start and we can build on it over time.

garrett:
Preview: https://garrett.github.io/cockpit-project.github.io/faq.html

garrett:
Squashed to just a few separate commits (CSS support, FAQ). Ready for review.

Parts of this probably _need_ editing.

martinpitt:
#586 landed, so the next time this can be rebased and drop the webp migration.

faq.md
```diff
@@ -0,0 +1,139 @@
...
+- this list will be replaced by the Table of Contents
+{:toc}
+
+## Installation
+
+### How do I install a recent version of Cockpit on Debian or Ubuntu?
+
+While Debian and Ubuntu both ship with access to Cockpit in their main repositories, the version in each might be older than the latest available. This is due to the nature of "stable" and "long term support" distributions.
+
+There's often a newer version of Cockpit available for both Debian and Ubuntu, in an extra "backports" repository. We recommend using this newer version, if you're able to do so.
```
martinpitt, line 14: Huh, since when did GitHub lose the ability to insert suggestions?

There's often a newer version of Cockpit available in the "backports" repositories, for both [Debian](https://backports.debian.org/) and [Ubuntu](https://wiki.ubuntu.com/UbuntuBackports).

faq.md
```diff
@@ -0,0 +1,139 @@
...
+
+Please read the installation instructions for the appropriate distribution:
+- [Debian installation instructions](running.html#debian)
+- [Ubuntu installation instructions](running.html#ubuntu)
+
+### How do I use Cockpit with a proxy?
```
martinpitt, line 20: with a reverse proxy

A normal "forward" proxy should be fine.

faq.md
```diff
@@ -0,0 +1,139 @@
...
+Read one of the following documents:
+
+- [Proxying Cockpit over NGINX](external/wiki/Proxying-Cockpit-over-NGINX.html)
+- [Proxying Cockpit over Apache with LetsEncrypt](external/wiki/Proxying-Cockpit-over-Apache-with-LetsEncrypt.html)
+
+## Development
+
+### How do I build an app for Cockpit?
+
+- [Starter Kiit](blog/cockpit-starter-kit.html)
```
martinpitt, line 33: "Kit"

faq.md
```diff
@@ -0,0 +1,139 @@
...
+- [Make your Cockpit page easily installable](blog/making-a-cockpit-application.html)
+- [Additional developer-related resourcesare on the Contributing page](external/wiki/Contributing.html)
```
martinpitt, line 35: Missing space in "resources are"

I'm a bit wary of repeating links here. But indeed the Contributing page is a bit of a mish-mash: it applies to both contributing to cockpit itself, as well as creating third-party pages for it.

faq.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Troubleshooting
+
+### Login
+
+#### Blank white page after login
```
martinpitt, line 41: If this happens, it's always a good first step to check both the browser console (Ctrl+Shift+J in most browsers) as well as the system logs (e.g. `sudo journalctl --since "5 minutes ago"`) for related error messages.

faq.md
```diff
@@ -0,0 +1,139 @@
...
+
+Newer updates of Safari no longer have this issue, but some older [devices may need to use a workaround](running/safari.html).
+
+### Software update
+
+#### Error message about "OSTree"
+
+You're seeing "OSTree is not available on this system" or "Unable to communicate with OSTree" instead of a software update page.
+
+In all liklihood, this means you have `cockpit-ostree` installed on a system that does not use `rpm-ostree` to manage its packages. You should be using Cockpit's built-in PackageKit-based software updater, not `cockpit-ostree`.
```
martinpitt, line 68: lik*e*lihood

.. should be using Cockpit's PackageKit-based software updater in `cockpit-packagekit`, not ...

faq.md
```diff
@@ -0,0 +1,139 @@
...
+    ```
+    nmcli con add type dummy con-name fake ifname fake0 ip4 1.2.3.4/24 gw4 1.2.3.1
+    ```
+
+3. Reboot
+
+##### Explanation
+{:.no_toc}
+
+Both Debian and Ubuntu use two different network stacks which don't play so well together.
```
martinpitt, line 108: Debian does not use netplan by default, and it's unlikely that people use it. People could explicitly configure networkd, though, but that's equally true on Fedora and all other distros. So I'd perhaps leave this out and restrict this section to Ubuntu/netplan?

faq.md
```diff
@@ -0,0 +1,139 @@
...
+
+Cockpit's software update page uses PackageKit, which checks networkd. However, as online status is managed by netplan on Debian and Ubuntuinstead, networkd reports back to PackageKit that the network is offline as a critical error message and stops further actions.
```
martinpitt, line 110: ... which checks NetworkManager. However, as the network interfaces are primarily managed by netplan and systemd-networkd in Ubuntu, NetworkManager reports back ...

faq.md
```diff
@@ -0,0 +1,139 @@
...
+
+##### Explanation
+{:.no_toc}
+
+Cockpit uses PackageKit for the Software updates page. PackageKit does not currently support excludes, except in `.repo` files.
+
+More details at:
+- https://bugzilla.redhat.com/show_bug.cgi?id=1237014#c6
+- https://github.com/cockpit-project/cockpit/issues/17955
+- https://www.systutorials.com/how-to-exclude-a-package-from-a-specific-repository-only-in-yum/
```
martinpitt, line 139: Let's make these clickable links.

martinpitt (changes requested):
Thank you! This should be a live document, so I dont' want to discuss possible other FAQs here. I just have a bunch of factual corrections and typos (a lot of places are missing spaces, is your keyboard kaputt? :grin: ).

I'd actually prefer this to be on the wiki instead of the homepage, to encourage some more community involvement. But as you already went that far, it's fine here as well.

Cheers!

faq.md
garrett:
```suggestion
- [Starter Kit](blog/cockpit-starter-kit.html)
```

faq.md
garrett:
```suggestion
- [Additional developer-related resources are on the Contributing page](external/wiki/Contributing.html)
```

faq.md
garrett:
```suggestion
There's often a newer version of Cockpit available in the "backports" repositories, for both [Debian](https://backports.debian.org/) and [Ubuntu](https://wiki.ubuntu.com/UbuntuBackports). We recommend using this newer version, if you're able to do so.
```

faq.md
garrett:
```suggestion
### How do I use Cockpit with a reverse proxy?
```

faq.md
garrett:
```suggestion
In all likelihood, this means you have `cockpit-ostree` installed on a system that does not use `rpm-ostree` to manage its packages. You should be using Cockpit's PackageKit-based software updater in `cockpit-packagekit`, not `cockpit-ostree`.
```

faq.md
garrett:
```suggestion
Cockpit's software update page uses PackageKit, which checks NetworkManager. However, as the network interfaces are primarily managed by netplan and systemd-networkd in Ubuntu, NetworkManager reports back to PackageKit that the network is offline as a critical error message and stops further actions.
```

faq.md
garrett: Good point!

garrett:
OK, I've updated all your suggestions and have updated the preview.

garrett:
Oh, we don't have a link to the FAQ from anywhere. 

(We can add this in a follow-up. Don't let the link block the FAQ.)

martinpitt (approved):
Very nice, thanks @garrett! Agreed that adding a link can be a follow-up.

PR #589, opened 2022-12-02
Don't repeat Cockpit client link in blog post

skobyda:
generate-release-notes script had Cockpit client link in "footer_dynamic", which was added to footer in loop for each released repo. This caused that footer would contain multiple of the same "Cockpit client" links.

[Cockpit 279 release notes](https://cockpit-project.org/blog/cockpit-279.html) are example of this:
Screenshot from 2022-12-02 12-27-27


martinpitt (approved):
Thanks!

PR #594, opened 2022-12-15
FAQ update

garrett:
Preview:

- FAQ updates on the FAQ: https://garrett.github.io/cockpit-project.github.io/faq.html

Additionally:

- Footer of https://garrett.github.io/cockpit-project.github.io/
- Bottom section of https://garrett.github.io/cockpit-project.github.io/documentation.html

### DIFF
--- a/documentation.md
+++ b/documentation.md
@@ -35,3 +35,9 @@ The [Cockpit guide](guide/latest/) contains:
 3. [Developer guide](guide/latest/development.html) contains topics such as embedding and integrating Cockpit, packages, and the Cockpit API.
 
 More [developer-focused resources are on the contribute page](external/wiki/Contributing.html).
+
+## Additional help
+
+- [Frequently Asked Questions (FAQ)](faq.html)
+- [Cockpit's forum @ GitHub Discussions](https://github.com/cockpit-project/cockpit/discussions/)
+- 
--- a/documentation.md
+++ b/documentation.md
@@ -40,4 +40,3 @@ More [developer-focused resources are on the contribute page](external/wiki/Cont
 
 - [Frequently Asked Questions (FAQ)](faq.html)
 - [Cockpit's forum @ GitHub Discussions](https://github.com/cockpit-project/cockpit/discussions/)
-- 
--- a/faq.md
+++ b/faq.md
@@ -5,6 +5,27 @@ title: Frequently Asked Questions (FAQ)
 - this list will be replaced by the Table of Contents
 {:toc}
 
+## General
+
+### How do I contact the Cockpit team?
+
+- If you're filing an issue, use our issue tracker.
+   - For general Cockpit bugs and feature requests, use [Cockpit's issue tracker](https://github.com/cockpit-project/cockpit/issues).
+   - For other parts of Cockpit use specific issues:
+      - [Website issues](https://github.com/cockpit-project/cockpit-project.github.io/issues)
+      - [Virtual machine related issues](https://github.com/cockpit-project/cockpit-machines/issues)
+      - [Podman UI related issues](https://github.com/cockpit-project/cockpit-podman/issues)
+      - [OStree related issues](https://github.com/cockpit-project/cockpit-ostree/issues)
+    - If you're not sure where to file the issue, [use the general issue tracker](https://github.com/cockpit-project/cockpit/issues). (We can move it if necessary.)
+    - If it's not an issue, ask us in the forum or IRC.
+- There's [a discussion forum for Cockpit on GitHub](https://github.com/cockpit-project/cockpit/discussions).
+- The Cockpit team is on IRC, usually during European business hours, from Monday through Friday, at #cockpit on irc.libera.chat. You need [a registered nick](https://libera.chat/guides/registration) in order to join.
+- [Cockpit has a mailing list](https://lists.fedorahosted.org/admin/lists/cockpit-devel.lists.fedorahosted.org/).
+
+### How do I create an application for Cockpit?
+
+All development-focused documentation is located on the [Contribute page](https://cockpit-project.org/external/wiki/Contributing.html) of the website. This includes creating your own [app page with the starter kit](blog/cockpit-starter-kit.html) and other related topics.
+
 ## Installation
 
 ### How do I install a recent version of Cockpit on Debian or Ubuntu?
@@ -16,6 +37,18 @@ There's often a newer version of Cockpit available in the "backports" repositori
 Please read the installation instructions for the appropriate distribution:
 - [Debian installation instructions](running.html#debian)
 - [Ubuntu installation instructions](running.html#ubuntu)
+
+### How do I use Cockpit on another port?
+
+Cockpit runs on port 9090 by default. If you'd like to change the port to something else, please [consult the Cockpit guide's section on TCP port and address](https://cockpit-project.org/guide/latest/listen).
+
+### How do I avoid opening a port on my production machines?
+
+It's possible to not run the Cockpit socket systemd service and log into your target machines with SSH.
+
+If you use a Linux-based desktop, use the [Cockpit Client desktop application, available on Flathub](https://flathub.org/apps/details/org.cockpit_project.CockpitClient).
+
+If you cannot use Cockpit Client, we provide Cockpit's web server in the [cockpit/ws container image](https://quay.io/repository/cockpit/ws). You can run that on any other host with SSH access to the target machine. This includes running the container on Kubernetes.
 
 ### How do I use Cockpit with a reverse proxy?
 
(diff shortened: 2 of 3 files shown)


jelly:
Screenshot from 2022-12-15 16-39-01

This seems to have an empty bullet point

jelly (changes requested):

documentation.md
```diff
@@ -35,3 +35,9 @@ The [Cockpit guide](guide/latest/) contains:
 3. [Developer guide](guide/latest/development.html) contains topics such as embedding and integrating Cockpit, packages, and the Cockpit API.
 
 More [developer-focused resources are on the contribute page](external/wiki/Contributing.html).
+
+## Additional help
+
+- [Frequently Asked Questions (FAQ)](faq.html)
+- [Cockpit's forum @ GitHub Discussions](https://github.com/cockpit-project/cockpit/discussions/)
+-
```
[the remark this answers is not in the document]
garrett: The important parts of the wiki are mirrored on the website, where they're integrated and searchable. So: No.

But if you think there's anything useful on the wiki that isn't on the website, we can adjust things a bit and mirror it on the website.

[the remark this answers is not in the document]
garrett:
```suggestion
```

faq.md
```diff
@@ -138,4 +174,25 @@ Cockpit uses PackageKit for the Software updates page. PackageKit does not curre
...
+
+On many servers, desktops, and laptops, the setting is turned off by default.
+
+##### Solution
+{:.no_toc}
+
+1. Turn off the computer that is running Cockpit. Then, turn it back on and immediately press the appropriate key to enter the BIOS / EFI settings (e.g. Delete, F1, F12, etc.).
+2. In the BIOS / EFI settings, look for the option to enable virtualization and ensure it is enabled.
+3. Save the changes and exit the BIOS / EFI settings.
+4. Wait for the computer to finish booting and try starting a VM in Cockpit again.
```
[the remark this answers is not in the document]
garrett: Yep, it's common enough. And not obvious. The best solution would be if we could check for it in Cockpit Machines, but right now we cannot, and older versions in long-term distros wouldn't get the fix even if we did have a way and added it today, so I think the question deserves to be answered.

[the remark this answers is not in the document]
jelly: I think we should create an issue in cockpit-machines so this at least get's fixed in the future then.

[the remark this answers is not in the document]
garrett: Done. https://github.com/cockpit-project/cockpit-machines/issues/891

faq.md
```diff
@@ -5,6 +5,27 @@ title: Frequently Asked Questions (FAQ)
...
+      - [OStree related issues](https://github.com/cockpit-project/cockpit-ostree/issues)
+    - If you're not sure where to file the issue, [use the general issue tracker](https://github.com/cockpit-project/cockpit/issues). (We can move it if necessary.)
+    - If it's not an issue, ask us in the forum or IRC.
+- There's [a discussion forum for Cockpit on GitHub](https://github.com/cockpit-project/cockpit/discussions).
+- The Cockpit team is on IRC, usually during European business hours, from Monday through Friday, at #cockpit on irc.libera.chat. You need [a registered nick](https://libera.chat/guides/registration) in order to join.
+- [Cockpit has a mailing list](https://lists.fedorahosted.org/admin/lists/cockpit-devel.lists.fedorahosted.org/).
+
+### How do I create an application for Cockpit?
+
+All development-focused documentation is located on [the Contribute page](https://cockpit-project.org/external/wiki/Contributing.html) of the website. This includes creating your own [app page with the starter kit](blog/cockpit-starter-kit.html) and other related topics.
```
martinpitt, line 27:
```suggestion
All development-focused documentation is located on the [Contributing page](https://cockpit-project.org/external/wiki/Contributing.html) of the website. This includes creating your own [app page with the starter kit](blog/cockpit-starter-kit.html) and other related topics.
```

faq.md
```diff
@@ -17,6 +38,12 @@ Please read the installation instructions for the appropriate distribution:
 - [Debian installation instructions](running.html#debian)
 - [Ubuntu installation instructions](running.html#ubuntu)
 
+### How do I use Cockpit on another port?
+
+Cockpit runs on port 9090 by default. If you'd like to change the port to something else, please [consult the Cockpit guide's section on TCP port and address](https://cockpit-project.org/guide/latest/listen).
+
+Additionally, it's possible to stop the Cockpit socket systemd service and log into Cockpit by SSH only. For this to work, you have to use a Linux-based desktop and run [the Cockpit Client desktop application, available on Flathub](https://flathub.org/apps/details/org.cockpit_project.CockpitClient).
```
martinpitt, line 45: This feels a little misplaced here. I'd make this another section, like "How do I run Cockpit without opening a new port?". We can then also mention the bastion mode of the cockpit/ws container.

faq.md
```diff
@@ -61,6 +88,15 @@ Older versions of Safari on iPhones, iPads, and Safari on macOS on M1 (ARM) mach
 
 Newer updates of Safari no longer have this issue, but some older [devices may need to use a workaround](running/safari.html).
 
+##### Blank page with incorrect time
+
+A blank page after logging in also may be a symptom of a wildly incorrect system time on your server, as all certficiates have a start and end date.
```
martinpitt, line 93: This sounds very unplausible to me, do you have a link where this happened? The TLS certificates are evaluated by the browser, not by the server. And if they expire, they are simply invalid. In neither case they appear as blank page, but the browser won't let you visit cockpit at all.

faq.md
```diff
@@ -138,4 +174,25 @@ Cockpit uses PackageKit for the Software updates page. PackageKit does not curre
...
 - [Cockpit bug #17955: Cockpit update is ignoring excludes in yum/dnf configurations](https://github.com/cockpit-project/cockpit/issues/17955)
-- [SysTutorials: How to exclude a package from a specific repository only in yum?](https://www.systutorials.com/how-to-exclude-a-package-from-a-specific-repository-only-in-yum/)
\ No newline at end of file
+- [SysTutorials: How to exclude a package from a specific repository only in yum?](https://www.systutorials.com/how-to-exclude-a-package-from-a-specific-repository-only-in-yum/)
+
+### Virtual machines
+
+#### Virtual machines not working as expected
+
+If a VM hangs during boot or you see an error message such as "No bootable device", it may be due to a configuration option in your computer's BIOS / EFI settings.
```
martinpitt, line 183: Again I don't understand this. The VM boot isn't related to the host's BIOS/EFI settings at all. Perhaps the single exception is KVM support, but depending on how the VM is configured, that will either just make it very slow, or not start at all (if the VM requires KVM). I have no proposal how to update the text, as I don't know what it means. Do you have a link to the original issue/report? (as a FAQ, it should have been asked at least twice, I figure :grin: )

martinpitt (dismissed):
Thanks! Happy to fix some things here, but for some I need your help.

faq.md
garrett: It's literally called "Contribute" in the navigation. I originally had it like your suggestion but realize that and changed it accordingly. But I'm fine with your suggestion too.

faq.md
garrett: Sure. Please feel free to split it off.

(I'd rather not block the FAQ updates too long, one way or another.)

faq.md
garrett: Yes, it happened. https://github.com/cockpit-project/cockpit/issues/17995

faq.md
garrett: This happens. Here's a recent issue: https://github.com/cockpit-project/cockpit-machines/issues/879

This is tied to a setting in BIOS/EFI and I've seen it myself before in the past too.

I always set that to enabled, but I have forgotten on a new system and have even seen it reset on a BIOS update (and ran into the not being able to start a VM issue as a result). Since the person who reported it and I have both experienced it (multiple times over the years), I think it's worth a mention.

garrett: I also opened an issue @ https://github.com/cockpit-project/cockpit-machines/issues/891 to add a check for it. However, even once we implement something like this (if we can) then it still doesn't cover Enterprise / LTS versions of Cockpit, so I think we still would want to mention it in the FAQ.

faq.md
martinpitt: Ah, I see -- the menu says "Contribute", the page title "Contributing". Not important enough. Including "the" in the link feels a bit strange, but again,  not important.

faq.md
martinpitt: But it's not clear whether the wrong time happened on the server or the client, and what browser was used. Why would the browser accept an invalid certificate for the html/css pages, but not for the websocket?  This wasn't asked "frequently", not debugged sufficiently, and sounds very dubious to me. IMHO we shouldn't duplicate all weird corner cases, otherwise the FAQ would both become really long and filled with dubious stuff, an also not really be "F".

I converted that issue into a discussion https://github.com/cockpit-project/cockpit/discussions/18060 and asked the reporter about some details.

faq.md
martinpitt: Thanks!

faq.md
martinpitt: I split it out into a new paragraph. WDYT?

martinpitt:
I pushed some updates as discussed above. I am happy with this now, but as I messed around in your PR, I'd rather have @jelly re-review this. Thanks!

martinpitt (approved):

martinpitt:
@jelly : Note that the preview in the discussion is still out of date. Please look at the diff here.

PR #597, opened 2022-12-19
Start using 'available' instead of 'included' when referring to cockpit

### LINKED ISSUE #596: Instructions state "Cockpit is included in Ubuntu 17.04 and later" but this is not true
### Explain what happens

Checking the instructions for Ubuntu it simply states "Cockpit is included in Ubuntu 17.04 and later" but it is not installed on any of my Ubuntu Server 20.04 or 22.04 machines or on my Ubuntu Desktop 22.04 machine. I would assume there is some caveat or exclusion somewhere, I have not specifically not installed cockpit. The instructions I think need clarifying slightly.

### Version of Cockpit

264-1ubuntu0.22.04.1

### Where is the problem in Cockpit?

Documentation

### Server operating system

Ubuntu

### Server operating system version

20.04 & 22.04

### What browsers are you using?

Firefox

### System log

_No response_

KKoukiou:
When referring to cockpit package being available through a distro's default package manager for installation, use the 'available' word and not 'included'.
Included is slightly misleading making users think that the package is installed by default in the relevant distro.

Closes #596

### DIFF
--- a/running.md
+++ b/running.md
@@ -85,7 +85,7 @@ However, we __strongly__ encourage you to use the latest version of your browser
 {% endcomment %}
 {% assign check = '<span aria-label="check">✓</span>' %}
 
-| |Tested|Included||
+| |Tested|Available||
 |-|-|-|-|-|-|{%
   for distro in site.data.distros
 %}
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Hmm.. I wouldn't understand "included in Debian" as "in a default install". But sure, thanks!

gh3597:
> Hmm.. I wouldn't understand "included in Debian" as "in a default install". But sure, thanks!

No harm in making it 100% clear though.

PR #614, opened 2023-03-20
scripts: Add two rewrite rules for HACKING

garrett:
Added support for rewriting the new relative links in HACKING.md for https://github.com/cockpit-project/cockpit/pull/18509.

### DIFF
--- a/_scripts/update-external-docs.rb
+++ b/_scripts/update-external-docs.rb
@@ -40,6 +40,10 @@ bots_pages = %w[
   'https://github.com/cockpit-project/cockpit/blob/master/HACKING.md' =>
   '{{ site.baseurl }}/external/source/HACKING.html',
 
+  './pkg/' => 'https://github.com/cockpit-project/cockpit/tree/main/pkg',
+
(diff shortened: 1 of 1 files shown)


_scripts/update-external-docs.rb
```diff
@@ -40,6 +40,10 @@
   'https://github.com/cockpit-project/cockpit/blob/master/HACKING.md' =>
   '{{ site.baseurl }}/external/source/HACKING.html',
 
+  './pkg/' => 'https://github.com/cockpit-project/cockpit/tree/main/pkg',
```
martinpitt, line 43: Is the left side an exact match or substring? You are replacing `pkg/` with `pkg`, so a `pkg/foo` would be substituted with `pkgfoo`. Probably safer to consistently append a slash or not.

martinpitt (changes requested):

martinpitt:
I triggered a new run, and https://github.com/cockpit-project/cockpit-project.github.io/pull/615/files looks good. https://cockpit-project.org/external/source/HACKING has working links.

PR #645, opened 2023-07-12
Add headscale and cloudflared applications

gbraad:
This adds two additional applications for headscale management and cloudflare tunnels overview. Both are in a very early state, but are usable. This also fixes the accidental override of the `sensors` id, which was therefore not shown. 🤦 copy-n-paste mistake.


martinpitt (approved):
Thank you! Headscale, Tailscale, I'm getting old :sweat_smile:

PR #650, opened 2023-07-28
Update faq.md

cmo-pomerium:
Adding Pomerium proxy's existing Cockpit guide as an option for Cockpit users.


martinpitt (approved):
Thanks!

PR #656, opened 2023-08-09
scripts: Handle redirects in image URLs

garrett:
GitHub changed the way they handle images; their URLs might not have an extension now. For the time being, we should assume the images without extensions are PNG files.

We will probably want to download whatever they are and run some identification on them in the future (like the `file` command).

I also switched the string extension and string concat to use native Ruby File.* commands.

### DIFF
--- a/_scripts/generate-release-notes.rb
+++ b/_scripts/generate-release-notes.rb
@@ -178,19 +178,29 @@ def build_frontmatter
   }.to_yaml.gsub(/^:/, '')
 end
 
-# Download the image from GitHub and save locally
-def download_image(url, basename)
-  extension = url.split('.').last
-  filename = "#{basename}.#{extension}"
+# Download the image from a GitHub and save locally
+def download_image(url, basename, limit = 3)
+  raise ArgumentError, 'too many HTTP redirects' if limit == 0
 
-  # Download and save file
-  image = Net::HTTP.get(URI(url))
-  File.write("images/#{filename}", image)
+  uri = URI(url)
+  response = Net::HTTP.get_response(uri)
 
-  @files_images.push("images/#{filename}")
+  case response
+  when Net::HTTPRedirection
+    new_url = response['location']
+    download_image(new_url, basename, limit - 1)
+  when Net::HTTPSuccess
+    extension = File.extname(uri.path)
+    local_file = "#{basename}#{extension}"
+    File.write("images/#{local_file}", response.body)
+    @files_images << "images/#{local_file}"
+  
+    # Return the local file on success
+    local_file
+  end
 
-  # Return filename (with extension)
-  filename
+rescue StandardError => e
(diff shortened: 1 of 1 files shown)


_scripts/generate-release-notes.rb
```diff
@@ -180,14 +180,17 @@ def build_frontmatter
 
 # Download the image from GitHub and save locally
 def download_image(url, basename)
-  extension = url.split('.').last
+  extension = File.extname(url)
+  # If an extension isn't specified, then default to PNG
+  extension = 'png' if extension.to_s.empty?
   filename = "#{basename}.#{extension}"
```
jelly, line 186: This now adds a duplicate `.` for png images.

_scripts/generate-release-notes.rb
```diff
@@ -180,14 +180,17 @@ def build_frontmatter
 
 # Download the image from GitHub and save locally
 def download_image(url, basename)
-  extension = url.split('.').last
+  extension = File.extname(url)
+  # If an extension isn't specified, then default to PNG
+  extension = 'png' if extension.to_s.empty?
```
jelly, line 185: This won't work:

```
[jelle@t14s][~/projects/cockpit-project.github.io]%curl https://github.com/cockpit-project/cockpit/assets/200109/bb33f362-a4ac-470e-a9b9-62ee5c3d3116.png
Not Found%
```

jelly (changes requested):
This does not seem to be fixable by this script, we can't get the images even if we hack the url:

Working old url:

https://user-images.githubusercontent.com/3228183/211269003-613885c7-6a11-4416-b69d-6435560ec49a.png

New url:

https://github.com/cockpit-project/cockpit/assets/200109/bb33f362-a4ac-470e-a9b9-62ee5c3d3116

Modify into user-images url just fails so *ugh* not much we can do.
```
curl https://user-images.githubusercontent.com/200109/bb33f362-a4ac-470e-a9b9-62ee5c3d3116.png
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>TRQ7PN4ARBTBNPFN</RequestId><HostId>x8JWBK0ueMfd6TzZK5GVAi4XiYSUGkr/mtve1V/z120GXG5ACa7z32fGDLVkL0B5PTnumegDPiY=</HostId></Error>%
```

_scripts/generate-release-notes.rb
jelly: Ok this just needs to be redirected!

garrett:
Addressed your issues and integrated your suggestions @ https://github.com/cockpit-project/cockpit-project.github.io/pull/657, but added multiple redirect support (recursive calling based on response) and error handling.

jelly (approved):
Would have liked to have a comment about the new urls which we need to handle now but let's get it merged. :+1:

PR #657, opened 2023-08-09
Handle new Github issue inline image urls

jelly:
Newly pasted inline images in Github issues now don't end on a .png and need to be redirected once to an aws image url.


jelly:
Superseeded by https://github.com/cockpit-project/cockpit-project.github.io/pull/656

PR #674, opened 2023-11-13
Update Ubuntu/Mint install instructions

martinpitt:
See individual commits for details.

Preview: https://martinpitt.github.io/cockpit-project.github.io/running.html#ubuntu

### DIFF
--- a/running.md
+++ b/running.md
@@ -240,11 +240,6 @@ We recommend installing or updating the latest version from backports. This repo
 sudo apt install -t ${VERSION_CODENAME}-backports cockpit
 ```
 
-On Linux Mint, use `UBUNTU_CODENAME` instead of `VERSION_CODENAME` like so:
-```
-sudo apt install -t ${UBUNTU_CODENAME}-backports cockpit
-```
-
 {:.note}
 When updating Cockpit-related packages and any dependencies, make sure to use `-t ...-backports` as above, so backports are included.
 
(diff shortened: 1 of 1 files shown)


running.md
```diff
@@ -231,7 +231,7 @@ When updating Cockpit-related packages and any dependencies, make sure to use `-
 {:.note}
 These commands require a POSIX compatible shell like `bash`. For other shells like `fish`, temporarily run `bash -i`.
 
-Cockpit is available in Ubuntu 17.04 and later, and available [as an official backport](https://help.ubuntu.com/community/UbuntuBackports) for 16.04 LTS and later.
+Cockpit is available in all current Ubuntu releases, and available [as an official backport](https://help.ubuntu.com/community/UbuntuBackports) for LTS releases.
```
allisonkarlitskaya, line 234: 16.04 is surely the last supported LTS release as well, no?  Could we say "all current Ubuntu releases, including all currently supported LTS releases"?

allisonkarlitskaya (approved):
Quite fine with this.  One question.

running.md
garrett: "Supported" could be interpreted as our own support, so I think we should avoid that, unless we find a clear way to say it. I agree that calling out LTS here would be redundant... and "current" would include LTS, IMO.

We could say "maintained" instead, or we could simplify it completely.

```suggestion
Cockpit is available in maintained Ubuntu releases, and available [as an official backport](https://help.ubuntu.com/community/UbuntuBackports) for LTS releases.
```

```suggestion
Cockpit is available in Ubuntu, with [updated versions in official backports for LTS releases](https://help.ubuntu.com/community/UbuntuBackports).
```

(I prefer something like this second suggestion. WDYT?)

martinpitt: I actually just thought "let's just say Ubuntu". So fully agreed, I'll take the second suggestion. Thanks!

PR #686, opened 2023-12-08
search.html: Add icon

### LINKED ISSUE #685: Search Icon on the Search Field
Page: 

Feature : Search Icon on search field

Steps to Replicate: 
Navigate https://cockpit-project.org/search.html
In the Search filed there be Search Icon

Attachment :
Screenshot_239

ItsDee19:
Added search icon in the search tab.

Fixes #685

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -5,12 +5,18 @@
 
 {% unless include.template == false %}
 <div class="search-page">
-
   <!-- Search form -->
   <form class="search-form" id="search-form" action="https://duckduckgo.com/" role="search">
-    <input autocomplete="off" id="search-input" name="q" class="search-input" aria-labelledby="searchFormButton">
-    <input type="hidden" name="sites" value="cockpit-project.org">
-    <button class="search-button" id="searchFormButton">Search</button>
+    <div class="search-container">
+      <input autocomplete="off" id="search-input" name="q" class="search-input" aria-labelledby="searchFormButton">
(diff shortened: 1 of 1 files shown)


ItsDee19:
@jelly Can you please merge this PR?

martinpitt (changes requested):
I pushed that patch to my fork for an easy preview: https://martinpitt.github.io/cockpit-project.github.io/search.html

[Before](https://cockpit-project.org/search.html) it looked like this, and adjusted itself to the window width:

image

With this PR it looks very garbled:

image

The input needs to grow, and the elements need to be aligned horizontally.

It's also a weird order: In at least Cockpit itself, the icon usually comes *before* the input line, like here on the Services page:

image

So this needs some work.

Thanks!

FTR: I'm not sure that this even *needs* an icon -- there's only the single input on the page anyway.

garrett (changes requested):
This should be done with CSS. Set the background of the input to the SVG icon and add padding to the left of the input which should be equal to the same as icon + padding (if the icon doesn't have padding itself) + space between icon and text.

(It shouldn't be an SVG icon in a container like this.)

Also: 
- Where does this icon come from? What's the license?
- Can we use a PatternFly icon instead, if this isn't from PF? https://www.patternfly.org/design-foundations/icons/#all-icons

martinpitt:
Please update *this* PR with changes. It is utterly confusing to open new ones for each iteration, as that destroys the discussion and reviews. Thanks!

PR #687, opened 2023-12-11
Update search.html latest

ItsDee19:
@garrett please check and merge PR.

### DIFF
--- a/_includes/search.html
+++ b/_includes/search.html
@@ -3,15 +3,46 @@
 <script src="{{ site.baseurl }}/assets/lib/search.js"></script>
 {% endunless %}
 
+  <style>
+      
+.search-form {
+  display: flex;
+  flex-direction: row;
+  justify-content: center;
+}
+
+form.nosubmit {
+ border: none;
+ padding: 0;
+}
+
+input.nosubmit {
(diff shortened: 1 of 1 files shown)


garrett (changes requested):
Thanks!

However:

1. This shouldn't need a wrapper in the HTML; the form is already a built-in wrapper for the elements
2. CSS should go in a CSS file (for example: `search.sass` already exists https://github.com/cockpit-project/cockpit-project.github.io/blob/main/assets/lib/search.sass)
3. CSS shouldn't use element name prefixes (it's overspecified and elements can change in the document even when classes might not)
4. `.nosubmit` doesn't make sense as a class name; it doesn't have anything to do with not submitting (and in fact, the form will submit a search to DuckDuckGo as a fallback, if JavaScript is not enabled)
5. `.search-form` has layout already (grid, but it _could_ be changed to flex) https://github.com/cockpit-project/cockpit-project.github.io/blob/main/assets/lib/search.sass#L3-L7
6. `.search-input` is where you'd want to add the background image https://github.com/cockpit-project/cockpit-project.github.io/blob/main/assets/lib/search.sass#L3-L7

PR #692, opened 2024-01-02
Convert styles to SCSS

garrett:
This rewrites the SASS files to be SCSS (the version of SASS that looks more like CSS, which is what we use across Cockpit, and therefore more familiar with anyone developing Cockpit or CSS).

It's intentionally in two commits, as one is for the platform libraries and the other is for the site-specific styles.

(This should change nothing visually for the website, but should improve the consistency of the developer experience.)


garrett:
I'll close this as-is and send up a new overhaul (with these changes and more) in a bit.

PR #695, opened 2024-01-08
search: Fix unstable sorting in Chrome

garrett:
While `return scoreA < scoreB` works in Gecko and WebKit and _used to_ work in Blink, sorting "broke" (it still worked, but not as intended) in Chrome at some point.

Switching the less-than comparison to a minus makes all browser engines work the same (as Gecko, WebKit, and an older Blink engine does in either variant).

Preview is at https://garrett.github.io/cockpit-project.github.io/search.html#javascript

### DIFF
--- a/assets/lib/search.coffee
+++ b/assets/lib/search.coffee
@@ -81,7 +81,7 @@ $ ->
         scoreB = pages[b].score = (titleB * 3 + summB + dateScoreB) / 5 / 0.71
 
         # Sort based on score
-        return scoreA < scoreB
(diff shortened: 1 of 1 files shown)


martinpitt (approved):
Thanks, good catch! `<` was just plain wrong, and it could only have worked half of the time (when you interpret `true` as `1`). The [API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) is supposed to return -1/0/0 for smaller/equal/bigger.

PR #696, opened 2024-01-09
running, data: Promote Arch to "tested"

garrett:
We test on Arch, so we should mention that on our running/install page.

I've added the tested flag and reordered the distros to keep all tested ones next to each other.


jelly (approved):
Thanks!

PR #699, opened 2024-01-10
style: Improve search on mobile

garrett:
Move padding of button into both button and input; allow the form to wrap on mobile.


garrett:
This is a super simple fixup (as I had ported search from grid to flex in the change and forgot to let it wrap); I'll go ahead and merge it.

PR #701, opened 2024-01-23
style: Add a default scroll margin

garrett:
This adds a little space above a fragment-linked item (foo.html#bar), for in-page linking.

I've added a --scroll-margin variable which can be used to customize it per page. But we rely on the fallback for now.

### DIFF
--- a/assets/site.scss
+++ b/assets/site.scss
@@ -147,6 +147,10 @@ body {
   display: none;
 }
 
+[id] {
+	scroll-margin-block-start: var(--scroll-margin, 2rem);
+}
+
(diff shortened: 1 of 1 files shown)


garrett:
Without:

image

With:

image

(There's just a little bit more space above elements with IDs that get ID targeting.)

This is especially useful for floating elements in page, like with sticky. (Which can be useful for Cockpit itself). But we're not using sticky here, so it's just a little more breathing room at the top of the page when fragment-linking inside page.

martinpitt (approved):
Thanks!

PR #703, opened 2024-01-30
fix: add Raspberry Pi required step

mlavi:
https://github.com/cockpit-project/cockpit/issues/8477#issuecomment-1824633520 discovered a missing dependency that completes the solution in the FAQ.

Confirmed on my Raspberry Pi 4b with Ubuntu 22.04.3 LTS (jammy), so I'm improving the FAQ.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -136,7 +136,13 @@ Create a placeholder file and network interface.
     nmcli con add type dummy con-name fake ifname fake0 ip4 1.2.3.4/24 gw4 1.2.3.1
     ```
 
-3. Reboot
+3. Skip this step unless you have a Raspberry Pi; install extra Linux kernel modules for networking:
+
+    ```
+    sudo apt install linux-modules-extra-raspi
+    ```
+
+4. Reboot
 
 ##### Explanation
 {:.no_toc}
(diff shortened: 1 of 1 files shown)


faq.md
```diff
@@ -136,7 +136,13 @@ Create a placeholder file and network interface.
     nmcli con add type dummy con-name fake ifname fake0 ip4 1.2.3.4/24 gw4 1.2.3.1
     ```
 
-3. Reboot
+3. Skip this step unless you have a Raspberry Pi; install extra Linux kernel modules for networking:
```
martinpitt, line 139: Let's avoid the double negative, and be more specific -- this applies to Ubuntu (and possibly Raspbian), but e.g. not Fedora. How about
```suggestion
3. If you run Ubuntu arm64 (e.g. on a Raspberry Pi), install extra Linux kernel modules for networking:
```

Also, this step must move to 2 -- you need the kmods *before* creating a dummy net device (current step 2).

mlavi: I agree about the double negative and better verbiage.
I had made the changes in that order originally and the reboot makes the effect the same, but I agree it couldn't hurt to move it earlier.
Thanks for your review, I'll incorporate or submit another PR if I can't figure it out here. :)

martinpitt (approved):
Cheers!

PR #706, opened 2024-02-02
Update cockpit metainfo

CtrlZmaster:
At some point, cockpit changed the application id and this page linked from documentation includes an outdated example. If used, this results in AppStream metainfo for the app being dropped as it doesn't extend any other app.


martinpitt (approved):
Ooh, thanks a lot!

PR #717, opened 2024-03-22
Update Cockpit applications metadata

### LINKED ISSUE #718: Unlisted application - Cockpit Identities
I recently discovered an application had been released and is in active development and support, by a 3rd party. However it hasn't been added to the list, this means its not on the website but also not listed in Cockpit either.

So to increase awareness of its presence I create a pull request, this issue is here to accompany it.

MrGrymReaper:
I recently found out an Cockpit application which is currently released and in active development, though not listed on the official website's applications list. Due to it being excluded from this list people are more likely to not know about this application, in order to fix this issue. I decided to contribute the necessary information to the project.

Fixes #718

### DIFF
--- a/_data/applications.yml
+++ b/_data/applications.yml
@@ -176,6 +176,15 @@ navigator:
   prerelease: false
   description: |
     A Featureful File Browser for Cockpit.
+
+identities:
+   title: Cockpit Identities
+   source: https://github.com/45drives/cockpit-identities
+   issues: https://github.com/45Drives/cockpit-identities/issues
+   official: false
+   prerelease: false
+   description: |
+     A Cockpit plugin to easily manage users and groups along with associated items.
  
 benchmark:
   title: Benchmark


_data/applications.yml
```diff
@@ -176,6 +176,15 @@ navigator:
...
     A Featureful File Browser for Cockpit.
+
+identities:
+   title: Cockpit Identities
+   source: https://github.com/45drives/cockpit-identities
+   issues: https://github.com/45Drives/cockpit-identities/issues
+   official: false
+   prerelease: false
+   description: |
+     A Cockpit plugin to easily manage users and groups along with associated items.
```
martinpitt, line 187: Can you please make the description a bit more concrete? The builtin Accounts page can do all the [listed features](https://github.com/45drives/cockpit-identities?tab=readme-ov-file#features) except for setting the Samba password. I.e. please drop "easily", if possible avoid "manage", and replace the vague "associated items" with something more specific. I.e. it should answer the question "why and when would I want to install this?" Unfortunately the upstream project itself isn't very verbose in that regard.

martinpitt (changes requested):
Thanks!

_data/applications.yml
martinpitt: To be more specific -- the 45drives apps are not very reliable. They have zero tests, do, well, "interesting" things with shell commands, don't fit the general Cockpit UI look and guidelines, etc. Presumably they did these for a specific product, but Cockpit is supported across a wide range of operating systems.

In other words, we don't want to recommend this page for standard Unix user management. If there's some "killer feature" which the Accounts page doesn't have, it should be pointed out as such.

Venefilyn:
Closing as the 45Drives Cockpit Identities is no longer being developed and has several outstanding issues

Issue #725, opened 2024-05-03
Error when running _scripts/container-jekyll


deadmeu:
I'm trying to run an instance of the website in an EndeavourOS VM but am getting the following error:
```
[test@test cockpit-project.github.io]$ _scripts/container-create
## Creating website container...
STEP 1/2: FROM ghcr.io/github/pages-gem
Trying to pull ghcr.io/github/pages-gem:latest...
Getting image source signatures
Copying blob f12c0bb1b56e skipped: already exists
Copying blob 7bb465c29149 skipped: already exists
Copying blob 49b40be4436e skipped: already exists
Copying blob c558fac597f8 skipped: already exists
Copying blob d13bbd9d0903 skipped: already exists
Copying blob 3624f0415588 skipped: already exists
Copying blob 1aec4714ae57 skipped: already exists
Copying blob 2b9b41aaa3c5 skipped: already exists
Copying blob 0abfbb302d12 skipped: already exists
Copying blob 5dd011617356 skipped: already exists
Copying blob 6409e148743c skipped: already exists
Copying blob 95eaf8e4f682 skipped: already exists
Copying blob 5d3f8fce2cd4 skipped: already exists
Copying blob 3cfb8ed124d4 skipped: already exists
Copying blob 7b1dcf51c52a skipped: already exists
Copying blob b46f0b255d80 skipped: already exists
Copying blob c2eb13e07edc skipped: already exists
Copying config 0fd5994aba done   |
Writing manifest to image destination
STEP 2/2: RUN bundle config path /src/site/.gem
--> Using cache a1ee55dc16a49a02f30aef26fc3a2b99e2cd03250e3d052ecb888bd17f367a14
COMMIT cockpit-website
--> a1ee55dc16a4
Successfully tagged localhost/cockpit-website:latest
a1ee55dc16a49a02f30aef26fc3a2b99e2cd03250e3d052ecb888bd17f367a14
Bundle complete! 4 Gemfile dependencies, 96 gems now installed.
Bundled gems are installed into `./.gem`
## Done!
## Run _scripts/container-jekyll
[test@test cockpit-project.github.io]$ _scripts/container-jekyll
[...]

PR #727, opened 2024-05-03
apps: Display title without trailing whitespace

deadmeu:
There is a whitespace character trailing the app's title in the https://cockpit-project.org/applications.html page. This appears to be due to the `app-title` text containing additional newline and whitespace characters:
image

I think this change should resolve the issue, though I am unable to test it due to #725.

I'm opening this PR in case the maintainers agree with the fix without requiring me to resolve the testing issue.


martinpitt (approved):
Thanks!

PR #734, opened 2024-05-28
Add security contact according to RFC 9116

JanZerebecki:
This hopefully makes it to: https://cockpit-project.org/.well-known/security.txt

See also: https://github.com/cockpit-project/cockpit/pull/20529


martinpitt:
I copied this commit to my fork, and https://martinpitt.github.io/cockpit-project.github.io/.well-known/security.txt exists now. What kind of tools make use of that?

This is blocked until SECURITY.md actually exists. (I'd also like to move/rename that, but let's discuss that in the Cockpit PR).

JanZerebecki:
It is primarily used by humans.

PR #736, opened 2024-06-04
fix(running.md): broken archlinux web package link

yuioto:
archlinux has merged extra and community, and now all community packages are in extra https://archlinux.org/news/git-migration-completed/


jelly (approved):
Woops! Thanks for catching!

PR #738, opened 2024-06-10
Add Cockpit-files as official Cockpit project

jelly:


garrett:
Looked great. However, the source URL was incorrect and it was lacking and issues file. I checked, I fixed, and I squash+merged. Thanks!

PR #740, opened 2024-06-11
applications: Drop cockpit-certificates

### LINKED ISSUE #739: Demote or remove Cockpit-Certificates?
Since we've deprecated and archived cockpit-certificates as no one from the community stepped up to maintain it, we should reflect that on the website.

The question is if we should outright remove it or keep it around.

If we keep it listed on the website, we probably want to:

1. Not include it in the list of official Cockpit extensions.
2. Have a "deprecated" badge?
3. Include a notice in the description about it being deprecated.

martinpitt:
It's unmaintained and archived.
    
Fixes #739


garrett (approved):
Thanks!

PR #783, opened 2025-01-29
Adding instructions to setup cockpit website locally in windows

Mash707:


README.md
```diff
@@ -17,6 +17,12 @@ For more details on Springboard, see [jekyll-springboard](https://github.com/gar
 1. Create the container and install the dependencies with:
    `_scripts/container-create`
 
+#### Instructions for Windows
+
+0. Install [WSL 2](https://learn.microsoft.com/en-us/windows/wsl/install): `wsl --install` (Default Linux Distribution is Ubuntu)
+1. Clone the repository to your linux filesystem `/home/<username>`.
+2. Install Podman: `sudo apt install -y podman`
```
garrett, line 24: Adding a bit of descriptions around the instructions

```suggestion
#### Additional first-steps for Windows

Building the Cockpit website is possible on Windows too, thanks to "Windows Subsystem for Linux" — although it requires a little bit of extra setup before using the standard Linux instructions:

0. Install [WSL2](https://learn.microsoft.com/en-us/windows/wsl/install): `wsl --install` (Ubuntu is the default distribution in WSL2)
1. Clone the repository to your Linux filesystem `/home/<username>`
2. Install Podman: `sudo apt install -y podman`
3. You can now follow the instructions for Linux, inside your WSL2 Linux environment
```

README.md
```diff
@@ -33,6 +39,13 @@ So, for instant rendering of local changes, you'd run:
 
 - `_scripts/container-jekyll -Il`
 
+_Note: For Windows, you might need to redirect the ports because the browser in Windows is not able to access the IP address from outside the container._
+
+```
+sudo iptables -t nat -A PREROUTING -p tcp --dport 4000 -j DNAT --to-destination 127.0.0.1:4000
+sudo iptables -t nat -A PREROUTING -p tcp --dport 35729 -j DNAT --to-destination 127.0.0.1:35729
+```
+
```
garrett, line 48: I don't think this is necessary... I think the problem is that Jekyll is bound to localhost by default with 127.0.0.1, and does not accept from all hosts. But we do already have `-H 0.0.0.0` (which is the same as `--host 0.0.0.0`).

WSL2 is supposed to forward ports automatically too, and it seems like it does. However, I also saw that it only does this for ports below 5000. (Or was it 50000?)

I made a PR that should hopefully fix WSL2 without having to do any firewall stuff @ https://github.com/cockpit-project/cockpit-project.github.io/pull/784 — could you check it please?

Also, I see that Jekyll's instructions say to access it at <http://0.0.0.0:4000> when running the script. Have you tried that specifically? And have you tried other things like <http://localhost:4000> ?

Mash707: I have tried working accessing http://0.0.0.0:4000 and http://localhost:4000 in my browser it was not working. Then I manually redirected the ports then the site was loading using http://localhost:4000. It was a workaround that I did so I don't if the same problem can occur in other windows machines. That is why I added it as a note.
image

garrett: Ah, OK. Weird that it's not working by default.

Is there an easy way to set it up permanently?

Mash707: My bad, I re did the setup again. Now I able to access the site using http://localhost:4000/ . Also the live reload port is also working using http://localhost:35729/ .
image
We dont require to redirect the routes. I am not able to understand why I ran into this problem in the first place. Apologies for that.

Mash707: I'll remove the unnecessary information.

Mash707: I did the WSL2 installation this month only(It was not there previously). I am also not able to understand how it started to work. All I did was create a new clone , ran the build instructions and it worked.

Mash707:
@garrett Are we good to go for the this PR?

README.md
```diff
@@ -17,6 +17,15 @@ For more details on Springboard, see [jekyll-springboard](https://github.com/gar
 1. Create the container and install the dependencies with:
    `_scripts/container-create`
 
+#### Additional first-steps for Windows
```
martinpitt, line 20: it's "first steps". But perhaps "Initial setup on Windows"?

README.md
```diff
@@ -17,6 +17,15 @@ For more details on Springboard, see [jekyll-springboard](https://github.com/gar
...
+
+Building the Cockpit website is possible on Windows too, thanks to "Windows Subsystem for Linux" — although it requires a little bit of extra setup before using the standard Linux instructions:
```
martinpitt, line 22: Let's streamline this a little bit. What about:

> Running the container is possible on Windows by running a podman container in [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/install):

README.md
```diff
@@ -17,6 +17,15 @@ For more details on Springboard, see [jekyll-springboard](https://github.com/gar
...
+
+0. Install [WSL2](https://learn.microsoft.com/en-us/windows/wsl/install): `wsl --install` (Ubuntu is the default distribution in WSL2)
```
martinpitt, line 24: Install some Linux distribution (Ubuntu is the default): `wsl --install`

README.md
```diff
@@ -17,6 +17,15 @@ For more details on Springboard, see [jekyll-springboard](https://github.com/gar
...
+1. Clone the repository to your Linux filesystem `/home/<username>`
```
martinpitt, line 25: This isn't specific to WSL the git repo has to be cloned everywhere (also, we sort of assume that developers have a basic familiarity with git). Also /home/username is irrelevant here -- it doesn't matter where you check it out. If you care about documenting the `git clone`, please put this below "Get started", but I'd just leave it out.

README.md
```diff
@@ -17,6 +17,15 @@ For more details on Springboard, see [jekyll-springboard](https://github.com/gar
...
+2. Install Podman: `sudo apt install -y podman`
```
martinpitt, line 26: This should rather move into "Setting up the container" above, where it has the `dnf` command. I.e. it could say "dnf..." or "apt .." depending on your distribution.

martinpitt (changes requested):
Thanks! Some suggestions for cleanups. Let's do one more round, and then I'm happy to land -- we can clean this up further in follow-ups too.

We can squash all commits and edit the message during landing. But if you already want to squash on your side, please do of course.

README.md
Mash707: Cloning the repo on the Windows file system like `users/mash` makes the development process much slower. While working from the root directory it is much faster. 
I did experience this when I first try to build the website on windows file system, live reloads and the build process were slower. While on the linux filesystem it was pretty fast.

Mash707: May be we could add a suggestion like "Development process is much faster when working from the linux filesystem"

martinpitt: Oh, of course -- it didn't even occur to me that you would do this in the Windows file system, if you run the container on the Linux side.

README.md
Mash707: Will go with "Initial setup on Windows" provides more clarity

README.md
Mash707: So should I keep the cloning part?

martinpitt (approved):
Thanks! Let's not overthink this, this is nice!

PR #784, opened 2025-01-30
scripts: Add custom livereload port, hopefully fixing WSL2

garrett:
This shouldn't affect Linux (aside from another port being used), and this port is not even used unless the script is called with a flag that enables livereload.

It has the benefit of making the livereload port closer to the server port too. (Although it does not really matter, outside of the port being <5000, which affects WSL2 apparently.)

### DIFF
--- a/_scripts/container-jekyll
+++ b/_scripts/container-jekyll
@@ -1,3 +1,14 @@
 #!/bin/bash -e
 
-"$(dirname "$0")"/container-run bundle exec jekyll server -H 0.0.0.0 "$*"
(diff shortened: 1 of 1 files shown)


Mash707:
The port `35729` is working perfectly fine in wsl 2. I also tested the live reload functionality and its working perfectly. So, I don't think this change is necessary.. However, if we want to keep it closer to the server port, we can have this change.

garrett:
Thanks for your feedback! I'll go ahead and close this then.

PR #798, opened 2025-04-17
scripts: Omit bodhi links in release notes

mvollmer:
I heard we had decided some time ago that we want to remove them, and I don't like filling them out, so...

### DIFF
--- a/_scripts/generate-release-notes.rb
+++ b/_scripts/generate-release-notes.rb
@@ -88,8 +88,6 @@ VERSIONS ARE available now:
 
 @footer_dynamic = "
 * [NAME Source Tarball](https://github.com/cockpit-project/REPO/releases/tag/VERSIONS)
-* [NAME Fedora 42](https://bodhi.fedoraproject.org/updates/?releases=F42&packages=REPO)
-* [NAME Fedora 41](https://bodhi.fedoraproject.org/updates/?releases=F41&packages=REPO)
 "
 
 ### Code below ###
@@ -196,7 +194,7 @@ def download_image(url, basename, limit = 3)
     local_file = "#{basename}#{extension}"
     File.write("images/#{local_file}", response.body)
     @files_images << "images/#{local_file}"
-  
+
     # Return the local file on success
     local_file
   end


mvollmer:
We also need to update the release process docs when we merge this. Remove step 12 from https://github.com/cockpit-project/cockpit/wiki/ReleaseProcess

martinpitt:
Hmm, I wasn't involved in that discussion (the only thing I remember is stopping the email notifications, we added mastodon instead). What's wrong with the bodhi links? They are nice for testing feedback, and indeed our updates get tested fast and well.

mvollmer:
> Hmm, I wasn't involved in that discussion

@Garrett was, I wasn't either.

> They are nice for testing feedback, and indeed our updates get tested fast and well.

Makes sense.

garrett:
We all had this discussion _ages_ ago... IIRC, you were both in it? But it was a long, long time ago.

I looked it up and the discussion with the conclusion was 2022-06-29, so it was quite a long time ago... and we never actually did anything about it since:

https://docs.google.com/document/d/1LSRoK0reaMh51LPKM8S731Uo7TsB_LUKwZK2o30eXeQ/edit?tab=t.0#heading=h.qth6xbh8w4j2

> * \[garrett\] Revise release notes format  
>   * We give preference to Fedora. Is that needed? It’s build system stuff; Fedora will get it by default soon after anyway.  
>   * Change the preamble? Change the footer?  
>     * Every two weeks, no “regularly”

martinpitt:
I like the links, and they make it easy to test new versions and give feedback. But if you all feel strongly about it, I won't veto, so instead of :-1: I'll just abstain.

mvollmer:
> I like the links, and they make it easy to test new versions and give feedback.

The argument from three years ago was that we shouldn't give preference to any distribution in our release notes, right?

@garrett, could you bring this up again in our weekly meeting?

I'll close this until then.

garrett:
From: https://docs.google.com/document/d/1LSRoK0reaMh51LPKM8S731Uo7TsB_LUKwZK2o30eXeQ/edit?tab=t.0

Key takeaway:

> Decision: Just keep the search links to bodhi; don’t update them to specific package sub-pages on bodi

Do we need to update the script in this case? It already links to searches by default.

I think this should mean that we don't manually update the links and keep it as searches? We'll figure this out for next release though, I suppose.

PR #804, opened 2025-06-05
Replace double quotes from title attribute with a single quote

jelly:
Jekyll isn't smart enough to escape or translate this so leaving double quote strings generates invalid html.


jelly:
> Sure, thanks! Does that by any chance fix the current corruption on https://cockpit-project.org/ ?

Yes, that's the idea, tested it locally.

PR #806, opened 2025-06-05
Add ROS 2 Diagnostics 3rd party application

hilary-luo:
Hello,

We (Clearpath Robotics) have developed a new 3rd party application that we would like to add to the list of available plugins. It is available from source or debian through the instructions in the indicated repo. Please let me know if there is anything else needed.

Thanks,
Hilary


hilary-luo:
@martinpitt My turn to apologize for the delay. Please let me know if any other changes are required.

Issue #810, opened 2025-06-12
Dark Mode for Docs provides readability issues


shantanoo-desai:
For browsers that are either set to Dark Mode Settings, the documentation pages are not readable due to font color choices / theme font color choices.

There is also no icon to switch between Light Mode and Dark Mode causing readability issues.

## Example
Snapshot for the [Package Manifest Section](https://cockpit-project.org/guide/latest/packages.html#package-manifest)

### Mozilla Firefox 139.0.4 (Dark Mode Settings) && Google Chrome 137.0.7151.70

![Image](https://github.com/user-attachments/assets/2c2ca744-c8a8-4c60-81ab-46609b550600)

Either providing a Mode switch to change the CSS of the website or adapting the font colors for the Dark Mode might help.

mvollmer:
@Venefilyn, are you interested in this?

PR #811, opened 2025-06-12
css: Inherit color for variablelist

Venefilyn:
We had components that got colors from components or something else than
our own code. This changes table, variablelist and others to inherit the
color from what we set higher up. Should fix dark mode issues

Fixes: #810


mvollmer (approved):
Thanks!

PR #802, opened 2025-06-04
Cockpit 340

Venefilyn:
- Cockpit 340
- Cockpit Podman 107
- Cockpit Files 22
- Cockpit Machines 334
- Cockpit OSTree 211

Signed-off-by: Freya Gustavsson <freya@venefilyn.se>

(this is the same as #801 but just applied again after mistakenly pushing it to main before)

PR #819, opened 2025-07-28
Verify links with Lychee

Venefilyn:
Verifies links through [Lychee](https://github.com/lycheeverse/lychee) within the project are alive and doesn't go to
outdated or otherwise dead links.

### DIFF
--- a/.github/workflows/verify-links.yml
+++ b/.github/workflows/verify-links.yml
@@ -1,0 +1,69 @@
+name: Verify links
+
+on:
+  push:
+    branches:
+      - main
+      - workflow/verify-links # TODO REMOVE
+  repository_dispatch:
+  workflow_dispatch:
+  schedule:
+    - cron: "08 08 * * 1"
+
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write # required for peter-evans/create-issue-from-file
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Restore lychee cache
+        uses: actions/cache@v4
+        with:
+          path: .lycheecache
+          key: cache-lychee-${{ github.sha }}
+          restore-keys: cache-lychee-
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          fail: false
+          # Exclude all private and local addresses in the check
+          # Also exclude domains that actively block GitHub to send requests
+          # Any Cockpit URLs it finds that have variables will be ignored too
+          args: |
+            --max-concurrency 1
+            --retry-wait-time 60
+            --base .
+            --skip-missing
+            --exclude-all-private
+            --exclude '^https://linux.die.net'
+            --exclude 'file:///'
+            --exclude 'domain.tld'
+            --exclude '^.*\{\{'
+            --exclude 'https://bodhi.fedoraproject.org/updates/cockpit-*'
+            --cache
+            --cache-exclude-status 400..=599
+            --max-cache-age 1d
+            -v
+            .
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Find the last open report issue
+        if: steps.lychee.outputs.exit_code != 0
+        id: last-issue
+        uses: micalevisk/last-issue-action@v2
+        with:
+          state: open
+          labels: link-checker
+
+      - name: Update or create issue report
+        if: steps.lychee.outputs.exit_code != 0 && steps.last-issue.outputs.has-found == 'true'
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: Link Checker Report
+          content-filepath: ./lychee/out.md
+          issue-number: ${{ steps.last-issue.outputs.issue-number }}
+          labels: link-checker
--- a/.github/workflows/verify-links.yml
+++ b/.github/workflows/verify-links.yml
@@ -1,22 +1,42 @@
 name: Verify links
 
 on:
-  push:
-    branches:
-      - main
-      - workflow/verify-links # TODO REMOVE
+  # push:
+  #   branches:
+  #     - main
+  #     - workflow/verify-links # TODO Remove before merging PR
   repository_dispatch:
   workflow_dispatch:
+  workflow_call:
+    inputs:
+      create_issue:
+        required: false
+        type: boolean
+        default: false
   schedule:
     - cron: "08 08 * * 1"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  linkChecker:
+  link_checker:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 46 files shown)


jelly (changes requested):

_posts/2025-06-04-cockpit-340.md
```diff
@@ -0,0 +1,73 @@
+---
```
jelly, line 1: This seems to be the wrong commit? We discussed 340 in https://github.com/cockpit-project/cockpit-project.github.io/pull/802 and merged https://github.com/cockpit-project/cockpit-project.github.io/commit/dba26a6f343b06785c386e69fdf586f05f1eb0af

I guess you still had this in your own main and that's why its in the pull request?

Venefilyn: Woops, yeah you're right. This should only be about the workflow

.github/workflows/verify-links.yml
```diff
@@ -0,0 +1,69 @@
...
+            --exclude 'https://bodhi.fedoraproject.org/updates/cockpit-*'
+            --cache
+            --cache-exclude-status 400..=599
+            --max-cache-age 1d
+            -v
+            .
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Find the last open report issue
+        if: steps.lychee.outputs.exit_code != 0
```
tomasmatus, line 55: The [summary page](https://github.com/cockpit-project/cockpit-project.github.io/actions/runs/16807287057?pr=819) shows that it found some errors - can you please explain when will lychee fail and make an issue to notify us and what kind of problems?

(Maybe I'm just missing the point of link checkers)

.github/workflows/verify-links.yml
```diff
@@ -0,0 +1,69 @@
...
+        with:
+          path: .lycheecache
+          key: cache-lychee-${{ github.sha }}
+          restore-keys: cache-lychee-
+
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          fail: false
```
tomasmatus, line 32: if I understand right this does not affect the following step and its check `if: steps.lychee.outputs.exit_code != 0`

.github/workflows/verify-links.yml
jelly: In Cockpit we have `tools/urls-check` which creates an issue if the link check fails. Thats what this PR attempts to also do.

We generally haven't used much external actions so I am wondering if we want too.

.github/workflows/verify-links.yml
```diff
@@ -0,0 +1,69 @@
...
+          # Exclude all private and local addresses in the check
+          # Also exclude domains that actively block GitHub to send requests
+          # Any Cockpit URLs it finds that have variables will be ignored too
+          args: |
+            --max-concurrency 1
+            --retry-wait-time 60
+            --base .
+            --skip-missing
+            --exclude-all-private
+            --exclude '^https://linux.die.net'
```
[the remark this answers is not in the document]
Venefilyn: IIRC it was 403 forbidden. Likely due to detecting it's from GitHub and just flatout declining it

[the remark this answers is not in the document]
jelly: Yaml allows comments so would be good to add a reason

.github/workflows/verify-links.yml
```diff
@@ -0,0 +1,69 @@
...
+            --cache-exclude-status 400..=599
+            --max-cache-age 1d
+            -v
+            .
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Find the last open report issue
+        if: steps.lychee.outputs.exit_code != 0
+        id: last-issue
+        uses: micalevisk/last-issue-action@v2
```
jelly, line 64: This one seems to be dead for 2 years without releases and I am slightly concerned about supply chain attacks here.

It essentially does [one query](https://github.com/micalevisk/last-issue-action/blob/main/src/core.ts#L23) which seems to be trivially rewritten in JavaScript in the workflow like we do for `dependabot` :
```
            try {
              await github.rest.issues.removeLabel({
                owner: context.repo.owner,
                repo: context.repo.repo,
                issue_number: context.issue.number,
                name: 'node_modules'
              });
            } catch (e) {
              if (e.name == 'HttpError' && e.status == 404) {
                /* expected: 404 if label is unset */
              } else {
                throw e;
              }
            }
```

But then `github.rest.issues` using [actions/github-script](https://github.com/actions/github-script?tab=readme-ov-file#comment-on-an-issue)

Getting the issue might be as simple as `octokit.rest.issues.listForRepo({ labels: ["link-checker"] })` and we always create an urls-check issue with a fixed label.

.github/workflows/verify-links.yml
```diff
@@ -0,0 +1,69 @@
...
+          state: open
+          labels: link-checker
+
+      - name: Update or create issue report
+        if: steps.lychee.outputs.exit_code != 0 && steps.last-issue.outputs.has-found == 'true'
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: Link Checker Report
+          content-filepath: ./lychee/out.md
+          issue-number: ${{ steps.last-issue.outputs.issue-number }}
```
jelly, line 78: What is no issue is found? Does this accept null?

.github/workflows/verify-links.yml
Venefilyn: We could use the hash for taking the release and update it with Renovate when there are new versions (if ever). Would at least ensure that the release itself is the one we expect.
```suggestion
        uses: micalevisk/last-issue-action@044e1cb7e9a4dde20e22969cb67818bfca0797be # v2
```

But yeah might be better to use something like that if it works

jelly:
> @jelly I updated the code to be a bit better. PTAL at the output and let me know if there is anything specific that stands out that needs to be changed with config 🙏
> 
> https://github.com/cockpit-project/cockpit-project.github.io/actions/runs/18015821705?pr=819

I really like the idea of link checking but I see that this PR now also pushes an artifact of the build website and then in another action check the links? Why can't we do this in a normal git checkout?

Venefilyn:
> > @jelly I updated the code to be a bit better. PTAL at the output and let me know if there is anything specific that stands out that needs to be changed with config 🙏
> > [cockpit-project/cockpit-project.github.io/actions/runs/18015821705?pr=819](https://github.com/cockpit-project/cockpit-project.github.io/actions/runs/18015821705?pr=819)
> 
> I really like the idea of link checking but I see that this PR now also pushes an artifact of the build website and then in another action check the links? Why can't we do this in a normal git checkout?

Splitting it out makes it cleaner as we just have to build once. The artifact should be in GitHub runner's cache too IIRC so basically takes no time for it to "download" the artifact.

Otherwise we'd have to combine build and publish into one which isn't as clean, and we have to add the same build step for link checker as well.

We can't check our source as-is because we construct URLs and pages on build-time. Like for categories, authors, anything internal with `site_url()/`.

PR #837, opened 2025-11-10
_scripts: rewrite generate release notes into Python

jelly:
Most of our team knows Python better then Ruby, so it will be easier to enhance edit and fix the script written in Python. This also removes the need to run it with a Ruby development container.

---

Follow ups:

* Add something like --bodhi-links which replaces the placeholder links with something which parses bodhi's json api.
* Validate if the given user is in _data/users.yaml

### DIFF
--- a/_scripts/generate-release-notes
+++ b/_scripts/generate-release-notes
@@ -8,18 +8,19 @@ from a list of GitHub repositories.
 import argparse
 import contextlib
 import datetime
+import json
+import os
 import re
 import sys
-from pathlib import Path
-from typing import Any
-import yaml
-import os
-import json
 import urllib
 import urllib.error
-from urllib.request import Request, urlopen
-from urllib.parse import urlparse
 from html.parser import HTMLParser
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlparse
+from urllib.request import Request, urlopen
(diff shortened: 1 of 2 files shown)


_scripts/generate-release-notes
```diff
@@ -0,0 +1,458 @@
...
+            headers["Authorization"] = f"token {token.strip()}"
+
+    try:
+        # Specify agent as some websites otherwise block requests
+        req = Request(url=url, headers=headers)
+        resp = urlopen(req)
+        return json.loads(resp.read())
+    except urllib.error.URLError as e:
+        print(f"Could not fetch data from the GitHub API ({e})")
+        return None
```
martinpitt: AIs tend to do that. But I strongly suggest to drop these kind of exception handlers. We'd rather have the script crash and fix it than silently ignoring errors and creating incomplete output.

martinpitt (changes requested):
Thanks! I didn't review the logic in detail, I assume you tested this. This is fairly readable though, let's just clean up the error handling.

Venefilyn:
Added a commit that fixes a few bugs, allows exceptions to be raised, and changed `--released` to `--increment`. @jelly WDYT?

PR #847, opened 2025-12-08
Add high availability cluster application

CtrlZmaster:
Hi, I think it's high time we added our application to the list ;)


jelly (approved):
Thanks!

PR #854, opened 2026-01-29
Cockpit 355

jelly:


martinpitt (approved):
LGTM, thanks! Let's see if I can get @allisonkarlitskaya to review it, too. But how about we also get the robot's opinion?

@Copilot can you please review the language? Please read the previous posts and keep the general style and level of formality.

PR #859, opened 2026-02-10
Minimize the need to reboot after applying changes to NetworkManager

EvelynOctavia:
Tested on a "somewhat" fresh install of Ubuntu 24.04 server with Cockpit 314. Yup, when the `10-globally-managed-devices.conf` normally doesn't exist, simply touching the file works just fine.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -123,11 +123,10 @@ The software update page shows "packagekit cannot refresh cache whilst offline" 
 
 Create a placeholder file and network interface.
 
-1. Create `/etc/NetworkManager/conf.d/10-globally-managed-devices.conf` with the contents:
+1. Create an empty file on `/etc/NetworkManager/conf.d/10-globally-managed-devices.conf`:
 
     ```ini
-    [keyfile]
-    unmanaged-devices=none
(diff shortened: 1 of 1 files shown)


faq.md
```diff
@@ -123,11 +123,10 @@ The software update page shows "packagekit cannot refresh cache whilst offline"
 
 Create a placeholder file and network interface.
 
-1. Create `/etc/NetworkManager/conf.d/10-globally-managed-devices.conf` with the contents:
+1. Create an empty file on `/etc/NetworkManager/conf.d/10-globally-managed-devices.conf`:
 
     ```ini
-    [keyfile]
-    unmanaged-devices=none
+    sudo touch /etc/NetworkManager/conf.d/10-globally-managed-devices.conf
```
martinpitt: I get the "restart NM", but a merely empty file won't do much?

martinpitt (commented):
Thanks!

faq.md
OrchidLuna: I was looking at https://askubuntu.com/a/1075112, and in my testing, it appears to be working.

PR #874, opened 2026-03-13
logo: Update favicon to 2021 logo

Venefilyn:
This updates the favicons to use the new logo and also uses the Cockpit
blue that we are used to.

Process for getting the inverted logo with a blue circle with a white
plane I've taken the inner circle of the Cockpit logo and made the plane
white on top of it. This way the outer circle which would've been there
is not creating unnecessary whitespace.

Visually would look like this as a tab

image

vs old/current

image


tomasmatus (approved):
sure

PR #878, opened 2026-03-23
flatpak: Update images for Flathub

Venefilyn:
Current images are outdated and show older design. This updates the
images from my machine running latest stable Cockpit Client on
localhost.

Caveat of this is that overview shows

> System manufacturer System Product name

Which is strange but ultimately something my machine reports, so no bug
within Cockpit. However, might raise eyebrows for those looking at the
images.

Images are found in
https://flathub.org/en/apps/org.cockpit_project.CockpitClient


allisonkarlitskaya (changes requested):
https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/quality-guidelines#reasonable-window-size mentions

> In order to ensure that text and interface elements are visible scaled down in the app store UI the window size should be 1000x700 pixels or smaller (2000x1400 for HiDPI).

... without defining what HiDPI means or how we might indicate that our images are HiDPI.

These new screenshots are more like 3000x2000, so they need to be scaled down some.  In general, I guess the idea is that we should make the window "smaller" in the sense of having a higher ratio of content to padding, with fonts as large as possible (it mentions scaling down) without scrollbars.

It would also be good to look into how we define if these are HiDPI and what the scaling factor was...

Venefilyn:
@allisonkarlitskaya I'm struggling to get an answer from Flathub about what they mean with HiDPI and what the scaling factor would be etc.

I have a 4K monitor with 166% scaling so that's why the images are big. Are you able to take smaller images then? I scaled down almost as much as I could in the images I took so that there are no scrollbars while all images being the same size

jelly:
Small nit, I think it would be better if we took screenshots of Fedora and not Bazzite :-) 

A virtual machine gives me "Model" `QEMU Standard PC (i440FX + PIIX, 1996)` which is a bit nicer then the standard bios info. Also you might consider not publishing your machine-id as that's unique per machine.

razzeee:
The images should be smaller, even for hdpi

Venefilyn:
> The images should be smaller, even for hdpi

@razzeee what is HiDPI supposed to be? It isn't super obvious from the docs what relative sizing it is or something like that :D

razzeee:
Aim for 2000x1400 and set scale to 2

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots

Venefilyn:
@allisonkarlitskaya okay so either we can dismiss the reasonable window size and just go for smallest possible size without introducing scrollbars, cause otherwise it's just impossible

Here is the 1x size of Cockpit Client at 1000x700
Screenshot From 2026-04-14 17-41-27
Screenshot From 2026-04-14 17-40-55
Screenshot From 2026-04-14 17-41-45

And 2x/HiDPI size 2000x1400
Screenshot From 2026-04-14 17-42-36
Screenshot From 2026-04-14 17-42-18
Screenshot From 2026-04-14 17-42-06

I'd vote for going for 2x, setting smallest res which would be 2496x1802, or if I were to take it without HiDPI it would be 1248x901. 
image

Venefilyn:
@allisonkarlitskaya We could also go with the default requested size we get for Cockpit Client

image

allisonkarlitskaya:
They all seem kinda fine to me.  Let's talk tomorrow?

images/flatpak/flatpak-screenshot-overview-dark.png
Venefilyn, line 1: Another one just to show dark theme. Can be added to the metadata file

Venefilyn:
@allisonkarlitskaya Images updated with white theme, 1x sizing, and as small as I can without introducing scrollbars. PTAL.

Also added one dark-theme image which we'll add to the metadata file

allisonkarlitskaya (approved):
Wow those old screenshots are dated.  Thanks!

PR #884, opened 2026-04-14
images: Include plain Cockpit SVG

Venefilyn:
We currently don't have the logo accessible on a website anywhere and
instead we can only get the logo from GitHub, lets add it to the website
so it is more accessible if you do not want the 'Cockpit' text next to
it.


jelly (approved):
Sure!

PR #899, opened 2026-06-08
scripts: Release notes for Fedora 43,44

Venefilyn:
Removes 42 as we no longer package to it. Includes 44 instead.


jelly (approved):
Thanks!

PR #905, opened 2026-06-25
PRs: Add preview workflow using surge

Venefilyn:
Adds a PR preview workflows that
- Create a build from PR itself without secrets
- Publishes build artifact to surge.sh
- Tears down surge preview on PR closure

Reason for a split between build and publish has to do with a workaround
for security where workflows that are triggered on a pull request cannot
access secrets. As the deploy workflow just publishes an artifact it can
have access to secrets. Note this only has to do with PRs from forks and
does not matter for PRs within the repo itself.

We still need to explicitly approve workflows to run on outside
contributors.

Before merging this a surge token needs to be added to secrets as `SURGE_TOKEN`.
Ask me for token to add and I'll send it as I can't access the env or secrets in this repo

### DIFF
--- a/.github/workflows/pr-deployment.yml
+++ b/.github/workflows/pr-deployment.yml
@@ -1,0 +1,37 @@
+name: Surge PR Preview - Deploy Stage
+
+on:
+  workflow_run:
+    workflows: ["Surge PR Preview - Build Stage"]
+    types:
+      - completed
+
+permissions:
+  pull-requests: write # Needed to comment on PRs
+
+jobs:
+  # Important - the job id:
+  # MUST be unique across all surge preview deployments for a repository as the job id is used in the deployment URL
+  # MUST be kept in sync with the job id of the teardown stage (this id is used by the surge-preview action to forge the deployment URL)
+  deploy:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
+
+    steps:
+      - name: Download built site
+        uses: dawidd6/action-download-artifact@v21
+        with:
+          workflow: ${{ github.event.workflow_run.workflow_id }}
+          run_id: ${{ github.event.workflow_run.id }}
+          name: pr-build-dist  # Must match the name from build workflow
+          path: _site/
+
+      - name: Deploy to Surge
+        uses: afc163/surge-preview@v1
+        with:
+          surge_token: ${{ secrets.SURGE_TOKEN }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          build: echo done
+          dist: _site
+          failOnError: true
+          teardown: false # Teardown is handled by the separate workflow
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-deployment.yml
```diff
@@ -0,0 +1,37 @@
...
+    types:
+      - completed
+
+permissions:
+  pull-requests: write # Needed to comment on PRs
+
+jobs:
+  # Important - the job id:
+  # MUST be unique across all surge preview deployments for a repository as the job id is used in the deployment URL
+  # MUST be kept in sync with the job id of the teardown stage (this id is used by the surge-preview action to forge the deployment URL)
```
mvollmer, line 15: Just from reading the yaml here I can not figure out how these requirements are implemented. I think it would be good to explain that here as well.  Is the job id maybe just the string "deploy" on the next line?

(I am super naive with yaml and github workflows, and if I had to change this file for some reason I wouldn't understand what I have to do concretely to not violate the requirements. If I can't change the "deploy" name on the next line, it would really help me to have that spelled out explicitly.)

.github/workflows/pr-teardown.yml
```diff
@@ -0,0 +1,21 @@
...
+
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  pull-requests: write # Needed to comment on PRs
+
+jobs:
+  deploy: # Must match the job ID from the deploy workflow
```
mvollmer, line 11: Right, got it. "deploy" is the job id.  Maybe I am just too ignorant...

mvollmer:
I have been hosting pre-views of my PRs on Github itself, as explained in the README in this repo. But this works only for a single PR at a time and is not very automatic. So this is better, thanks!

mvollmer:
I would like @allisonkarlitskaya to approve this and install the token.

Venefilyn:
> I have been hosting pre-views of my PRs on Github itself, as explained in the README in this repo. But this works only for a single PR at a time and is not very automatic. So this is better, thanks!

It also doesn't work for me. Likely, both this repo and the forks of you and others got the deprecated deployment version enabled and I just can't. There isn't a way for me to deploy from a fork apart from like building locally and pushing to a gh-pages branch

.github/workflows/pr-deployment.yml
Venefilyn: I made changes since this was added I wonder if I can remove it.. I'll double-check

Venefilyn:
@allisonkarlitskaya I experimented with using GitHub itself and it would become very complex to get it to work properly. Essentially if we want fork previews it's better if they are hosted standalone from GitHub Pages as they can only host one site per repo. 

The workaround for this single-host is to have folders for PRs that are uploaded to the repo and hosted automatically, so in that case we could access `https://cockpit-project.org/pr-preview/905` for example. And for that to work we need to use `gh-pages` branch. But there is a problem with that as we need to handle both normal builds and separate PR builds to that branch in a secure way while also having the PR context. 

Using the approach of a split workflow like I've done here is the only way one could get GitHub Pages based approach to work for forks. But getting the PR number in that second workflow is not easy to do, but I did find a solution for it
https://github.com/orgs/community/discussions/25220#discussioncomment-8697399

As a note, I added a `SURGE_TOKEN` secret to this repo already but will experiment more with the GH Pages approach.

PR #908, opened 2026-06-30
PRs: Add preview workflow using GitHub Actions

Venefilyn:
Adds a PR preview workflows that
- Create a build from PR itself without secrets
- Publishes build artifact to preview PR subdirectory in gh-pages branch
- Removes preview PR subdirectory from gh-pages branch

Reason for a split between build and publish has to do with a workaround
for security where workflows that are triggered on a pull request cannot
access secrets. As the deploy workflow just publishes an artifact it can
have access to secrets. Note this only has to do with PRs from forks and
does not matter for PRs within the repo itself.

PR deployment and PR teardowns are run from the base repository and is
therefore secure.

Build workflows will need to be manually approved from outside
contributors. They will only be deployed if build workflow has
succeeded.

Upon PR preview being created a comment will be added to the PR itself
with a link to the preview URL. If this PR is run against our main we
will publish the preview under <cockpit-project.org/preview-pr/> and if
the PR is run against a fork it will publish the preview under
<username.github.io/cockpit-project.github.io/preview-pr/>. If the user
has a custom domain for their <username.github.io> the links will still
work but just be redirected by GitHub.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,70 @@
+name: Build Cockpit site
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
+# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
+concurrency:
+  group: "pages"
+  cancel-in-progress: false
+
+env:
+  REPO_OWNER: "${{ github.repository_owner }}"
+  REPO_NAME: "${{ github.event.repository.name }}"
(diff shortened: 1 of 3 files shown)


Venefilyn:
@allisonkarlitskaya Here it is.

This likely breaks deployment currently. I think we'd have to get rid of the GitHub build and deploy workflow altogether cause I'm running into this issue. If merged this would break URLs completely in its current state

- https://github.com/jekyll/github-metadata/issues/186

The thing with the GitHub workflow is also that it is extremely outdated, we're talking years old with no update on the workflow as well as a lot of the plugins

PR #913, opened 2026-07-23
Add ctop and manifest to applications.yml

ismetozalp:
Add two Cockpit plugins to the applications list:

- ctop (ismetozalp/ctop) — a btop-style system and process monitor for the Cockpit console: per-core CPU braille graphs, memory/disk, network, GPU, sensors, containers, and a live sortable process table with kill/signal, tree view, and Cockpit deep-links.

- manifest (ismetozalp/manifest) — a download manager for the Cockpit console: torrents, magnets, and HTTP/FTP/Metalink files driven by a per-user aria2c daemon, with a unified table view, per-file torrent selection, and 13 live themes.


jelly (approved):
Thanks

PR #915, opened 2026-08-11
Add `.well-known/security.txt`

Venefilyn:
Takes the .well-known security.txt file from redhat.com as that is the
appropriate reporting mechanism. While we do have the SECURITY.md file
in GitHub we shouldn't redirect here using our browser redirect as it
wouldn't work properly. Until we get redirects support we need to
populate this ourselves somehow.

Making it easy, we can grab Red Hat's security.txt as we have the same
reporting form. As a bonus it means the message is also signed by
Red Hat as nothing was changed.

Fixes: https://github.com/cockpit-project/cockpit-project.github.io/pull/734
Related-to: https://www.redhat.com/.well-known/security.txt


allisonkarlitskaya (approved):
- https://security.access.redhat.com/data/meta/v1/security.txt
- https://en.wikipedia.org/wiki/Security.txt

good enough for me

